PR #4, opened 2024-01-09
Refactor config + project ingestion

mawildoer:
## Why?

- Support multiple boards for the same project
- Make CI on the user-side easier

See: https://toil.kitemaker.co/2hAmxC-atopile/dSlNKj-atopile/items/48

## What?

- Add `mfg-data` target which controls the KiCAD CLI
- Supports multiple builds with varying entry-points
- Add a `-v` / `--verbose` option to the CLI, allowing users to say `-v` or `-vv` for increaing logging verbosity. This replaces `--debug`
- `--debug` option now starts a `debugpy` session and waits for connection
- Add `atopile-kicad` container which installs both, and has an easy-to-use entry point for CI 
- Build targets now receive a `BuildContext` object containing data they'd otherwise need to infer or derive from the config/cli/etc... individually

## Validation

I used the logic-card project to smoke the changes: https://github.com/Timot05/logic-card/pull/1/checks

## Related PRs

https://github.com/atopile/project-template/pull/4

### DIFF
--- a/src/atopile/cli/build.py
+++ b/src/atopile/cli/build.py
@@ -3,17 +3,16 @@
 import logging
 import shutil
 from functools import wraps
-from pathlib import Path
 from typing import Callable, Optional
 
 import click
-from attrs import frozen
 
-from atopile import address
-from atopile.bom import generate_bom as _generate_bom
-from atopile.bom import generate_designator_map as _generate_designator_map
-from atopile.cli.common import project_options, check_compiler_versions
-from atopile.config import Config
+import atopile.bom
+import atopile.front_end
+import atopile.netlist
+import atopile.manufacturing_data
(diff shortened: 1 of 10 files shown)


src/atopile/cli/build.py
```diff
@@ -27,71 +26,40 @@
 
 @click.command()
 @project_options
-@click.option("--debug/--no-debug", default=None)
 @muffle_fatalities
-def build(config: Config, debug: bool):
+def build(build_ctxs: list[BuildContext]):
     """
     Build the specified --target(s) or the targets specified by the build config.
```
Timot05, line 32: Would be nice to be able to build a specific target. For example `ato build default`.

Timot05 (approved):

PR #6, opened 2024-01-13
Cloud component db

napowderly:
Components are now fetched from ato component API!

Screenshot 2024-01-12 at 6 08 13 PM

Pretty simply, we just send all the data we have about the component, and the server finds a part using the specified parameters and returns a JSON with the part data.


mawildoer (changes requested):
There's a few comments that got orphaned on Gitlab as we moved over to Github.
Would you be able to address those before we merge this?https://gitlab.atopile.io/atopile/atopile/-/merge_requests/127

src/atopile/components.py
```diff
@@ -81,66 +71,112 @@ class NoMatchingComponent(errors.AtoError):
...
+
+def clean_cache():
+    """Clean out entries older than 1 day."""
+    for addr, entry in list(component_cache.items()):
+        cached_timestamp = datetime.fromtimestamp(entry["timestamp"])
+        if datetime.now() - cached_timestamp >= timedelta(days=1):
+            del component_cache[addr]
+    save_cache()
+
+import logging
```
mawildoer, line 135: These should probably be at the top of the file - they're currently casuing linting failures

src/atopile/front_end.py
```diff
@@ -75,10 +76,12 @@ def __repr__(self) -> str:
         return f"<{self.__class__.__name__} {self.address}>"
 
 
+from attrs import define, field
+
 @define
 class Physical(Base):
     """Let's get physical!"""
-    unit: pint.Unit
+    unit: pint.Unit = field(converter=lambda u: str(u) if isinstance(u, pint.Unit) else u)
```
mawildoer, line 84: This is typed as a `Pint` unit. It should be converted to a string like this on ingestion. Whatever is using this should do the conversion to a string - or, if this does warrant changing, the typing needs changing too

src/atopile/front_end.py
```diff
@@ -100,6 +103,18 @@ def tolerance_pct(self) -> float:
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__} {self.min_val} to {self.max_val} {self.unit}>"
 
+    def to_json(self) -> dict:
```
mawildoer, line 106: nit; this should be named `to_dict` since it's returning a dict, not a JSON string.

src/atopile/cli/create.py
```diff
@@ -3,8 +3,8 @@
 import re
 import sys
 import textwrap
-import webbrowser
 from pathlib import Path
+import webbrowser
```
mawildoer, line 7: nit; (wouldn't bother except there's two other linting errors blocking the PR)
Run "Organise Imports" command in VSCode, I believe this shouldn't have moved (and can impact merges if someone else has touched imports)

src/atopile/components.py
```diff
@@ -81,66 +55,111 @@ class NoMatchingComponent(errors.AtoError):
     title = "No component matches parameters"
 
 
-@cache
-def _get_generic_from_db(component_addr: str) -> dict:
+# Define the cache file path
+repo = Repo(".", search_parent_directories=True)
```
mawildoer, line 59: I think the comment I left on Gitlab still applies here, it makes the CLI inconsistent. Basically, this means your CWD needs to be inside the project you want to work on and you can't just say `ato build path/to/project/file.ato...`

src/atopile/components.py
```diff
@@ -81,66 +55,111 @@ class NoMatchingComponent(errors.AtoError):
...
-    if pd.isna(idx_min):
-        return filtered_df.iloc[0].to_dict()
+        lcsc = best_component["lcsc_id"]
+        log.debug(f"Successfully fetched component {lcsc} for {name}")
+        return best_component
 
-    return filtered_df.loc[idx_min].to_dict()
+    except requests.RequestException as e:
+        log.warning(f"API request failed: {e}")
+        return {"LCSC Part #": "Part not found", "value": "N/A", "unit": "N/A"}
```
mawildoer, line 162: I'd generally recommend using an `attrs` object or [`dataclass`](https://docs.python.org/3/library/dataclasses.html) for things like this. It's really easy to get dict keys wrong and type checkers can't help you with it.

src/atopile/front_end.py
```diff
@@ -78,7 +78,10 @@ def __repr__(self) -> str:
 @define
 class Physical(Base):
     """Let's get physical!"""
-    unit: pint.Unit
+
+    unit: pint.Unit = field(
```
mawildoer, line 82: re-adding this comment; I don't think we should be casting this to a string, or, if there's a reason we should be doing so, we need to also change the type-hint.

src/atopile/components.py
```diff
@@ -81,66 +71,112 @@ class NoMatchingComponent(errors.AtoError):
...
-            f"{value_range.unit} cannot be converted to {generic_unit} for $addr", addr = component_addr,
-            title="Invalid unit",
-        ) from ex
-
-    filters.append(f"min_value >= {min_float_val}")
-    filters.append(f"max_value <= {max_float_val}")
-
-    # Ensure the component's footprint is correct
-    filters.append(f"Package == '{_generics_to_db_fp_map[specd_data['footprint']]}'")
+        log.info(payload)
```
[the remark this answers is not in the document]
napowderly: Done

src/atopile/components.py
```diff
@@ -81,66 +71,112 @@ class NoMatchingComponent(errors.AtoError):
...
+        response = requests.post(url, json=payload)
+        response.raise_for_status()
+        best_component = response.json().get('Best Component')
```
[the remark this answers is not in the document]
napowderly: Fixed

src/atopile/components.py
```diff
@@ -81,66 +71,112 @@ class NoMatchingComponent(errors.AtoError):
...
-        raise NoMatchingComponent(msg, addr=component_addr)
+        if not best_component:
+            raise NoMatchingComponent("No valid component found", addr=component_addr)
 
-    # FIXME: Currently our cost function is dumb - it only knows dollars
-    # In the future this cost function should incorporate other things the user is
-    # likely to care about
-    idx_min = filtered_df["Price (USD)"].idxmin(skipna=True)
+        lcsc = best_component["lcsc_id"]
+        log.info(f"Successfully fetched component {lcsc} for {name}")
```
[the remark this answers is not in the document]
napowderly: fixed

src/atopile/cli/create.py
napowderly: fixed

src/atopile/components.py
napowderly: Updated

src/atopile/components.py
napowderly: Done

mawildoer (approved):

PR #7, opened 2024-01-13
Improve error in case of a build that doesn't exist

mawildoer:
## The error now looks like this:

Screenshot 2024-01-13 at 16 31 57

### DIFF
--- a/src/atopile/cli/common.py
+++ b/src/atopile/cli/common.py
@@ -30,6 +30,7 @@ def project_options(f):
     @click.option("-t", "--target", multiple=True)
     @click.option("-o", "--option", multiple=True)
     @functools.wraps(f)
+    @errors.muffle_fatalities
     def wrapper(
         *args,
         entry: str,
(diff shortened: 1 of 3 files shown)


Timot05:
Glad we are catching the error. The problem is actually that at the config does exist. Except that it is named `card-or`, which, probably because of the `-` can't be found.

mawildoer:
> Glad we are catching the error. The problem is actually that at the config does exist. Except that it is named `card-or`, which, probably because of the `-` can't be found.
> 
> Edit: actually it is found but changed to snakecase in the CLI

I thhhink I see the error and can fix it, but the cards are named with underscores in my PR targeting logic-cards: https://github.com/Timot05/logic-card/pull/2/files#diff-3f1bf466d147a54ef00ed389c120316add121cd7ad04c68e0258ef83dcbc1481R5

Mind sharing steps to reproduce the issue? Were you renaming the builds?

Timot05:
Can't invoke the `card-or` build: https://github.com/Timot05/logic-card/blob/main/ato.yaml

mawildoer:
> Can't invoke the `card-or` build: https://github.com/Timot05/logic-card/blob/main/ato.yaml

Oh, I see! You added it to main. Sorry I was still working off my PR.

Added a commit to fix it.

Timot05 (approved):
Tested and works

PR #10, opened 2024-01-15
Return early in case of missing semver tags

mawildoer:
## Smoked it with the same command in the issue:
Screenshot 2024-01-16 at 10 44 21


Timot05:
Did we actually want to allow people to install modules that lack semver? Aren't we better off not installing the module since it might lead to harder to track error once installed?

mawildoer:
> Did we actually want to allow people to install modules that lack semver? Aren't we better off not installing the module since it might lead to harder to track error once installed?

It's really convenient for development to allow people to install whatever.
If we didn't, they couldn't develop modules in the context of their project

Timot05 (approved):

PR #14, opened 2024-01-17
Resolve incorrect reference in build context error message

hollinsky:
It looks like the error message was meant to reference the other variable. Underlying problem (detecting `_autosave-*` and `~_autosave-*`) will still exist.


napowderly (approved):
Good catch, thanks!

PR #21, opened 2024-01-24
Improve error handling and robustness of cloud component access

mawildoer:


src/atopile/cli/build.py
```diff
@@ -147,4 +147,4 @@ def clone_footprints(build_args: BuildContext) -> None:
 
     for component in all_components:
         log.debug("Cloning footprint for %s", component)
-        download_footprint(component, dir=build_args.build_path / "build/footprints/footprints.pretty")
+        download_footprint(component, footprint_dir=build_args.build_path / "build/footprints/footprints.pretty")
```
napowderly, line 150: is build path already in /build? I think this should be "footprints/footprints.pretty"

napowderly (changes requested):
Sweet, thanks for cleaning that up!

src/atopile/config.py
```diff
@@ -173,12 +173,12 @@ def match_user_layout(path: Path) -> bool:
 @define
 class BuildContext:
     """A class to hold the arguments to a build."""
+    project_context: ProjectContext
```
mawildoer, line 176: Finish moving project-context to that variable

napowderly (approved):

PR #35, opened 2024-01-28
Top level only Netname algorithm

jarrod89:
see also #34

### DIFF
--- a/src/atopile/nets.py
+++ b/src/atopile/nets.py
@@ -10,7 +10,6 @@ from atopile.instance_methods import (
     all_descendants,
     get_children,
     get_links,
-    get_parent,
     iter_parents,
     match_interfaces,
     match_pins_and_signals,
@@ -63,24 +62,22 @@ class _Net:
 
     def generate_base_net_name(self) -> Optional[str]:
         """TODO:"""
-        WEIGHT_NO_GRANDPARENTS = 10
-        WEIGHT_INTERFACE_GRANDPARENT = 5
-        WEIGHT_SIGNAL = 2
 
         name_candidates = defaultdict(int)
-
+        min_depth = 100
         for signal in filter(match_signals, self.nodes_on_net):
-            name = get_name(signal)
-            if get_parent(signal) is None:
-                name_candidates[name] += WEIGHT_NO_GRANDPARENTS
-            elif any(map(match_interfaces, iter_parents(signal))):
-                name_candidates[name] += WEIGHT_INTERFACE_GRANDPARENT
-            else:
-                name_candidates[name] += WEIGHT_SIGNAL
+            min_depth = min(min_depth, len(list(iter_parents(signal))))
+        for signal in filter(match_signals, self.nodes_on_net):
+            # lower case so we are not case sensitive
+            name = get_name(signal).lower()
+            # only look at highest level
+            if min_depth == len(list(iter_parents(signal))):
+                name_candidates[name] += 1
 
         if name_candidates:
             highest_rated_name = max(name_candidates, key=name_candidates.get)
             self.base_name = highest_rated_name
+            #print(name_candidates)
 
 
 def _find_net_names(nets: Iterable[Iterable[str]]) -> dict[str, list[str]]:
--- a/src/atopile/nets.py
+++ b/src/atopile/nets.py
@@ -11,6 +11,7 @@ from atopile.instance_methods import (
     get_children,
     get_links,
     iter_parents,
+    get_parent,
     match_interfaces,
     match_pins_and_signals,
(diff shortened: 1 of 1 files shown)


napowderly:
Screenshot 2024-01-28 at 10 29 05 AM
Seems like we need to disregard shitty signal names like 'p1'. power.vcc would definately be a better name in this case

jarrod89:
I agree, that's the corner case I was talking about on discord. We could
prioritize interfaces over any other signal.. Although interfaces are not
always good. For example you have an interface in the generic capacitor so
this would make any net with a capacitor on it always take vcc or gnd, that
sucks.
The behavior of this PR isn't terrible. You could fix your circuit by
creating a gnd signal
Maybe we could have a regex filter for 'shit names' and create components
that follow this standard

napowderly:
We could just send the list of names to GPT4 and see what we should name it lol

jarrod89:
Youre joking about GPT4, but seriously, one of the requirements should be
good determinism. Should be able to explain to someone exactly how a name
is selected and how to force a name.

napowderly:
Indeed, I think we probably should give more context in most cases. P1 is obviously a bad name, micro.bypass_cap.p1 is at least orienting.

jarrod89:
It is supposed to do that when it finds a duplicate name, but there must be
some bug because it never does

mawildoer:
Ran a test of these changes on Tim's logic card project and ended up with this list:
`k vcc gnd adj vcc-1 gnd-1 gnd-2 wire not or and xor out vcc-2 xin xout swclk swdio run usb_dm usb_dp io3 sck io0 io2 io1 cs gpio0 gpio6 gpio7 gpio8 gpio9 gpio10 gpio11 gpio12 gpio13 gpio14 gpio15 gpio16 gpio17 gpio18 gpio19 gpio20 gpio21 gate gate-1 gate-2 gpio25 gpio26_a0 gpio27_a1 gpio28_a2 gpio29_a3 gnd-3" out-1 gnd-4" dp dm gnd-5" vcc-3 gnd-6" sub1 cc1 sub2 cc2 in wire-1 not-1 or-1 and-1 xor-1 latch out-2 r_sink g_sink b_sink r g b`

I'm not the ideal judge of whether this serves as an improvement or not.

Removed the debugging print statements, so happy to merge at your discretion!

jarrod89:
> Ran a test of these changes on Tim's logic card project and ended up with this list: `k vcc gnd adj vcc-1 gnd-1 gnd-2 wire not or and xor out vcc-2 xin xout swclk swdio run usb_dm usb_dp io3 sck io0 io2 io1 cs gpio0 gpio6 gpio7 gpio8 gpio9 gpio10 gpio11 gpio12 gpio13 gpio14 gpio15 gpio16 gpio17 gpio18 gpio19 gpio20 gpio21 gate gate-1 gate-2 gpio25 gpio26_a0 gpio27_a1 gpio28_a2 gpio29_a3 gnd-3" out-1 gnd-4" dp dm gnd-5" vcc-3 gnd-6" sub1 cc1 sub2 cc2 in wire-1 not-1 or-1 and-1 xor-1 latch out-2 r_sink g_sink b_sink r g b`
> 
> I'm not the ideal judge of whether this serves as an improvement or not.
> 
> Removed the debugging print statements, so happy to merge at your discretion!

Lots of gnds, what's going on there? I guess there are not many signals specifically called out.
The parent prepending function would be good here, have you looked in to why that's broken?

Timot05:
I reckon that when applying the net naming algorithm, having interfaces in the net that are not connected to anything is a problem. Simple example:

pin 1 ~ signal GPIO_1
GPIO_1 ~ signal pwm

And let's assume that nothing is actually connected to pwm in the rest of your system. I was just placed there for convenience. What name do you give to that net? "GPIO_1" or "pwm"? By not understanding what is actually on the path from one pin to other pin, we cannot take this decision reliably which should cause net naming issue.

jarrod89:
> I reckon that when applying the net naming algorithm, having interfaces in the net that are not connected to anything is a problem. Simple example:
> 
> pin 1 ~ signal GPIO_1 GPIO_1 ~ signal pwm
> 
> And let's assume that nothing is actually connected to pwm in the rest of your system. It was just placed there for convenience. What name do you give to that net? "GPIO_1" or "pwm"? By not understanding what is actually on the path from one pin to other pin, we cannot take this decision reliably which should cause net naming issue.

I don't understand your example, but I agree with what you said about unconnected interfaces being a problem, that's the cause of your gnd-1 thru 6. every capacitor has an interface, and its usually not used.
Maybe the answer is to not write ato code that way? maybe every capacitor shouldn't have a power interface.

PR #49, opened 2024-01-31
getfile accomodate Windows paths

IpsumDominum:


src/atopile/address.py
```diff
@@ -25,7 +25,9 @@ def get_file(address: AddrStr) -> str:
     This is because an "empty" file path is a valid address,
     to the current working directory, which is confusing.
     """
-    return address.split(":")[0]
+    #This is to ensure everything before the last colon
+    #Since Windows uses : in paths
+    return ":".join(address.split(":")[:-1])
```
mawildoer, line 30: Ah damn! Thanks for the PR!

Unfortunately in this case this fix won't work, because addresses `/path/to/file.ato:Path.To.Entry::path.to.instance`

The second double-colon fill break this patch.

It's been a little since I've used a windows but is it just the the `C:\...` at the beginning of the path?

IpsumDominum: Haha tbh I don't know either... I think it's just one : in the front, since you cannot create directories with : in them. 

But to be safe, would this work?
```
address_dirname,address_filename = os.path.split(address)
    return os.path.join(address_dirname,address_filename.split(":")[0])
```

mawildoer: Also a good idea! I gave something else a try if you wouldn't mind taking it for a spin?

Continues to work on OSx/presumably POSIXy things and the basics pass in a test! 😄

IpsumDominum: Seems to work for my one test case :)

There's also these below which needs fixing v 
Should I leave these to you?

```
def get_entry_section(address: AddrStr) -> Optional[str]:
    """
    Extract the root path from an address.
    """
    try:
        return address.split(":")[1]
    except IndexError:
        return None

def get_instance_section(address: AddrStr) -> Optional[str]:
    """
    Extract the node path from an address.
    """
    try:
        return address.split(":")[3]
    except IndexError:
        return None
```

This takes me to at least the next error.
`KeyError: 'Generics are missing data - internal error'`

```
def get_entry_section(address: AddrStr) -> Optional[str]:
    """
    Extract the root path from an address.
    """
    try:
        if len(address) >= 2 and address[1] == ":" and address[0].isalpha():
            return address[address.find(":")+1:].split(":")[1]
        return address.split(":")[1]
    except IndexError:
        return None


def get_instance_section(address: AddrStr) -> Optional[str]:
    """
    Extract the node path from an address.
    """
    try:
        if len(address) >= 2 and address[1] == ":" and address[0].isalpha():
            return address.split(":")[4]
        return address.split(":")[3]
    except IndexError:
        return None
```

PR #90, opened 2024-02-06
Reviewed and revised README.md for spelling and clarity

mofosyne:
Revised and fixed typos.

Also added some dot points to more quickly communicate the design intent of this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Then just `pip install atopile` and you're good to go!
 
 ## ❓ Why Atopile?
 
-The main overall objective of Atopile is to help push forward these pardiagrams from the software world to EDAs, mainly these points:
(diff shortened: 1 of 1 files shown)


mofosyne:
On a side note, wonder how well this would integrate with KiCAD

mawildoer:
Thanks for the PR!

mofosyne:
No probs.

Also regarding this point

> Continuous Integration (CI): Implement CI to guarantee high-quality, compliant designs with every commit, represented by a green checkmark for assurance.

Do you mean that this would support something like [SPICE ("Simulation Program with Integrated Circuit Emphasis")](https://en.wikipedia.org/wiki/SPICE) . I could imagine having a module be run though a SPICE simulation and checked against constants and expected behavioral outputs.

PR #172, opened 2024-02-21
Inspect improvements

Timot05:


src/atopile/cli/inspect.py
```diff
@@ -32,68 +33,59 @@
...
     """
-    visited = set()  # Keep track of visited nodes to avoid revisiting them.
-    connected_components = []  # Store the connected components.
+    def __init__(self, net: list[list[AddrStr]]):
+        self.inspect_net: list[AddrStr] = net
+        self.inspect_consumer: list[AddrStr] = []
+        self.context_net: list[AddrStr] = []
+        self.context_consumer: list[AddrStr] = []
+
+def find_nets_new(links: list[Link]) -> list[list[AddrStr]]:
```
mawildoer, line 48: nit; it's generally better to leave things like "new" out of names like this. Has little meaning and you don't have "find_nets" in this scope already. In this case, I'd probably also make it private (prefix it with `_`)

src/atopile/cli/inspect.py
```diff
@@ -32,68 +33,59 @@
...
+            if target not in graph:
+                graph[target] = []
+            graph[source].append(target)
+            graph[target].append(source)
+
     def dfs(node, component):
-        """
-        Perform depth-first search recursively to find all nodes connected to the current node.
-        """
         visited.add(node)
```
mawildoer, line 81: I'm perhaps missing them but this seems to be using variables declared beneath itself?

src/atopile/cli/inspect.py
```diff
@@ -255,28 +242,35 @@ def inspect(build_ctxs: list[BuildContext], inspect: str, context: Optional[str]
...
-    sorted_entries = natsort.natsorted(displayed_entries, key=lambda obj: obj.connectables[0])
-    for entry in sorted_entries:
-        pins = sorted(list(filter(match_pins, entry.connectables)))
-        signals = list(filter(match_signals, entry.connectables))
-        _add_row(pins, signals, list(set(entry.intermediate_connectable)), list(set(entry.consumers)))
+    #sorted_entries = natsort.natsorted(displayed_entries, key=lambda obj: obj.connectables[0])
+    for entry in inspect_entries:
+        pins = sorted(list(filter(match_pins, entry.inspect_net)))
+        signals = list(filter(match_signals, entry.inspect_net))
+        #interface = list(filter(match_interfaces, entry.inspect_net))
```
mawildoer, line 272: nit; remove commented out lines

mawildoer (approved):
Couple of nits, but if it's working then we're happy!

PR #239, opened 2024-03-10
Break dependency links by default

mawildoer:
https://toil.kitemaker.co/2hAmxC-atopile/dSlNKj-atopile/items/110


napowderly:
Screenshot 2024-03-10 at 5 34 13 PM
Ran ato install esp32-s3

mawildoer:
Did you try just the once? This doesn't seem related and I wonder if there was a delay to the server or something instead?

PR #255, opened 2024-03-15
Napowderly/fix signal name uniqueness

### LINKED ISSUE #222: Overriding variable names should crash the compiler
The below situation doesn't raise an error

Screenshot 2024-03-07 at 22 24 18

napowderly:
fixes #222

eg:
Screenshot 2024-03-15 at 2 19 08 PM

Screenshot 2024-03-15 at 2 18 22 PM


mawildoer (approved):
Approving since it's an upgrade, but it's worth noting that this one fixes one of the two conflict cases

PR #260, opened 2024-03-19
Add some debug info to uniqueness check

napowderly:


mawildoer (approved):

mawildoer (changes requested):
Actually, this will be formatted slightly strangely. It should be the address rather than the instance itself

napowderly:
> Actually, this will be formatted slightly strangely. It should be the address rather than the instance itself

Screenshot 2024-03-19 at 11 21 29 AM

napowderly (commented):
Modified to show address

napowderly:
> Actually, this will be formatted slightly strangely. It should be the address rather than the instance itself

Fixed

mawildoer (approved):

PR #267, opened 2024-04-01
Zone copy with net

### LINKED ISSUE #216: KiCAD extension should copy zones as well

nickkrstevski:
Fix #216 

**This deserves a thorough squiz**, seems hacky but setting zone net was tricky. **There must be a cleaner way to map source netname to target netname.**

- Duplicates source zone to target board
- Updates the net of the zone to new net name on target board
  - Finds net name of source zone
  - iterates thru source footprint pads until netname match is found
  - uses matched footprint uuid and uuid map to find target footprint
  - uses matched pad index of matched footprint to get new net information
  - Sets zone net to target net info
  - Sets zone to netcode 0 if no match found (no net)

### DIFF
--- a/src/atopile/kicad_plugin/common.py
+++ b/src/atopile/kicad_plugin/common.py
@@ -100,7 +100,6 @@ def update_zone_net(source_zone: pcbnew.ZONE, source_board: pcbnew.BOARD, target
                 new_netinfo = target_fp.Pads()[matched_pad_index].GetNet()
                 target_zone.SetNet(new_netinfo)
                 return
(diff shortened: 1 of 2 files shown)


mawildoer:
Perhaps this is information we can add the manifest that `atopile` generates on build. Currently it's just the source-target UUID map, but we could totally add a source-target net-name map too!

Side note, great PR Nick! Thank's for the step-by-step operation here. Makes it super easy to rip through!

napowderly:
Screenshot 2024-04-01 at 11 55 15 AM

Neat! Working for me on Kicad8

PR #272, opened 2024-04-02
Fix pushgroup bug and ref designators

### LINKED ISSUE #213: KiCAD extension should move reference designator position and size
The designators stay where they come by default and it sucks. Most of the time they're ages away from the component and 🤮

nickkrstevski:
Fixes pushgroup bug in `sync_tracks`
- After duplicating track from source board, ghost track remains on source board/group
- Doesnt happen in pull because source board is closed and not saved
- Exponentially adds tacks to source and target after multiple pushes
- Solved by removing ghost track from source board after duplication

Overloaded `find_anchor_footprint` function to allow board/group interchangability for both push/pull

Fix #213
- Copies over reference designator relative position to footprint
- Copies attributes(thickness, width, height, rotation)

### DIFF
--- a/src/atopile/kicad_plugin/pushgroup.py
+++ b/src/atopile/kicad_plugin/pushgroup.py
@@ -21,6 +21,7 @@ class PushGroup(pcbnew.ActionPlugin):
         source_board: pcbnew.BOARD = pcbnew.GetBoard()
         board_path = source_board.GetFileName()
         known_layouts = get_layout_map(board_path)
+        target_board = None
 
         # Push Selected Groups
         for g in source_board.Groups():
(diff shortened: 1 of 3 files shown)


nickkrstevski:
Added warning if no groups are selected. If you are actively editing group and try to push, the group is not selected according to KiCAD.

src/atopile/kicad_plugin/pushgroup.py
```diff
@@ -48,15 +55,29 @@ def Run(self):
...
+            # Shift all objects of interest by offset
+            for fp in target_board.GetFootprints():
+                fp.Move(offset)
+
+            for track in target_board.GetTracks():
+                track.Move(offset)
 
-        # Save the target board
-        target_board.Save(target_board.GetFileName())
+            # Save the target board
```
mawildoer, line 77: I indented this so it saved every target board you might've been updating, rather than just the first

mawildoer:
Thanks Nick!! 🥳

PR #273, opened 2024-04-03
Make KiCAD plugin ref syncing compatible with 8.0

mawildoer:
The `SetPos0` method no longer exists in the KiCAD 8 API.


src/atopile/kicad_plugin/common.py
```diff
@@ -134,7 +134,10 @@ def sync_footprints(
         target_fp.SetLayer(source_fp.GetLayer())
         # Ref Designators
         target_fp.Reference().SetAttributes(source_fp.Reference().GetAttributes())
-        target_fp.Reference().SetPos0(source_fp.Reference().GetPos0())
+        target_ref: pcbnew.FP_TEXT = target_fp.Reference()
```
mawildoer, line 137: @nickkrstevski perhaps there was a reason you didn't use `SetPosition` for this in the first place?

PR #279, opened 2024-04-10
Windows path fix

nickkrstevski:
Tiny PR - 1 character

The kicad plugin script `atopile.py` plugin_path is created with backslashes, this is no bueno for windows.

Switching to raw string fixed for windows, should work on unix.

Will this happen elsewhere?

### DIFF
--- a/src/atopile/cli/configure.py
+++ b/src/atopile/cli/configure.py
@@ -109,7 +109,7 @@ def install_kicad_plugin() -> None:
     """Install the kicad plugin."""
     # Find the path to kicad's plugin directory
     plugin_loader = f"""
-        plugin_path = "{Path(__file__).parent.parent}"
+        plugin_path = r"{Path(__file__).parent.parent}"
         import sys
         import importlib
 


src/atopile/cli/configure.py
```diff
@@ -109,7 +109,7 @@ def install_kicad_plugin() -> None:
     """Install the kicad plugin."""
     # Find the path to kicad's plugin directory
     plugin_loader = f"""
-        plugin_path = "{Path(__file__).parent.parent}"
+        plugin_path = r"{Path(__file__).parent.parent}"
```
mawildoer, line 112: We could even take it one further and just cast is to a string!

```suggestion
        plugin_path = str(Path(__file__).parent.parent)
```

Thank you Nick!!!

nickkrstevski: image

Won't this still treat the `\n` as an escape character?

mawildoer: Nah, it won't, but by pulling it to test the idea I realised what was wrong with it; we're dumping this whole contents to a file anyway!

You're on the money with this fix already

Thank you Nick!

mawildoer (approved):

PR #280, opened 2024-04-10
Add DRCs to KiCAD build

### LINKED ISSUE #265: DRC errors should fail builds
DRC checks should be run as part of CI. Errors like opens/shorts should fail the build.

mawildoer:
Closes #265


napowderly:
Screenshot 2024-04-11 at 3 19 25 PM
Not sure what the warning is about? But otherwise seems to work. Failed the first time due to a timeout after 5 seconds but has worked since then and I havent been able to replicate.

mawildoer:
That was expected, but also failed its purpose to tell you DRCs failed. Updated to this:
Screenshot 2024-04-11 at 18 00 39

PR #283, opened 2024-04-12
renames step model path to local build dir when footprints are copied in

napowderly:


src/atopile/cli/build.py
```diff
@@ -136,6 +136,17 @@ def consolidate_footprints(build_args: BuildContext) -> None:
         except shutil.SameFileError:
             log.debug("Footprint %s already exists in the target directory", fp)
 
+    """Post-process all the footprints in the target directory."""
+    for fp in fp_target.glob("**/*.kicad_mod"):
+        with open(fp, "r+", encoding="utf-8") as file:
+            content = file.read()
+            content = content.replace("{build_dir}", str(fp_target_step))
```
mawildoer, line 143: Where is this string being embedded in the first place?

src/atopile/cli/build.py
```diff
@@ -136,6 +136,17 @@ def consolidate_footprints(build_args: BuildContext) -> None:
...
+            file.seek(0)
```
mawildoer, line 144: This is a little weird; what's going on here?

src/atopile/cli/build.py
mawildoer: The answer is in easyeda2ato

src/atopile/cli/build.py
mawildoer: Looks like we're just re-writing this file without re-opening it with write permissions. Nothing too wild

PR #300, opened 2024-04-24
Add CD to the viewer

Timot05:


.github/workflows/cd.yml
```diff
@@ -33,6 +33,18 @@ jobs:
         with:
           python-version: '3.11'
 
+      -
+        name: Install Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '19.5.0' #TODO: check
```
mawildoer, line 40: TODO

.github/workflows/cd.yml
```diff
@@ -33,6 +33,18 @@ jobs:
...
+
+      -
+        name: Build Web Application
+        run: |
+          cd src/atopile/viewer
```
mawildoer, line 45: It'd be better to use the `working-directory` option instead of the `cd` command https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsworking-directory

(Since otherwise you'll need to `cd` back out again to not affect downstream steps)

pyproject.toml
```diff
@@ -112,6 +112,9 @@ exclude = [
     "docs/",
 ]
 
+# [tool.hatch.metadata.hooks]
```
mawildoer, line 115: Remove dead content

mawildoer (approved):

PR #318, opened 2024-04-26
Fix negative unaries - Take 2

mawildoer:
This is the same fix as #314 , except it should actually work...


src/atopile/parser/AtopileParser.g4
```diff
@@ -84,17 +84,12 @@ arithmetic_expression
     ;
 
 term
-    : term ('*' | '/') factor
-    | factor
+    : term ('*' | '/') power
+    | power
     ;
 
-factor
```
mawildoer, line 91: This will mean that you can't put a unary in front of brackets

src/atopile/parser/AtopileParser.g4
```diff
@@ -112,13 +107,11 @@ arithmetic_group
...
     | bilateral_quantity
-    | implicit_quantity;
+    | quantity;
 
-bound_quantity: quantity_end 'to' quantity_end;
-quantity_end: NUMBER name?;
-bilateral_quantity: bilateral_nominal PLUS_OR_MINUS bilateral_tolerance;
-implicit_quantity: NUMBER name?;
-bilateral_nominal: NUMBER name?;
+bound_quantity: quantity 'to' quantity;
```
mawildoer, line 112: Do we want to make tolerances (especially the `to` operator) part of the nesting arithmetic expressions rather than a part of an atom

PR #346, opened 2024-05-09
change signal/pin name format

napowderly:
change signal/pin format from:
name: anode/1

to:

signal: anode
pin: 1


mawildoer (changes requested):
This PR contains a lot of unrelated changes. It needs rebasing

PR #360, opened 2024-05-13
Save component positions

Timot05:
You can now place your components on the canvas and their positions will be saved in the ato-lock.yaml file. 

Screenshot 2024-05-13 at 14 44 35


src/atopile/cli/view.py
```diff
@@ -43,9 +43,10 @@ async def send_viewer_data():
 
 class Pose(BaseModel):
     """The position, orientation, flipping etc... of an element."""
-    x: float
-    y: float
-    angle: int  # degrees, but should only be 0, 90, 180, 270
+    position: dict[str, float] # {x: 0.0, y: 0.0}
```
Timot05, line 46: Not as strict enforcement as it was (since we don't know the contents of the dict). Hopefully that is fine?

src/atopile/schematic_utils.py
```diff
@@ -210,6 +223,21 @@ def get_schematic_dict(build_ctx: atopile.config.BuildContext) -> dict:
...
+def get_ato_lock_file(build_ctx: atopile.config.BuildContext) -> dict:
+    ato_lock_contents = {}
+
+    if build_ctx.project_context.lock_file_path.exists():
+        with build_ctx.project_context.lock_file_path.open("r") as lock_file:
+            ato_lock_contents = yaml.safe_load(lock_file)
+
+    return ato_lock_contents
+
+def get_pose(ato_lock_contents: dict, id: str) -> tuple[dict[str, float], float, bool, bool]:
```
mawildoer, line 235: I'd ideally pull the `Pose` definition into a `viewer_core` or `viewer_common` file, so you can return the right types in all the right places

Timot05: Done

PR #398, opened 2024-06-19
feat(kicad): update kicad plugin to support pull/push of DRAWINGS

### LINKED ISSUE #397: KiCAD plugin doesn't pull / push Drawings and other documentation / layers for modules
I have a design which builds using various modules I've defined. I have defined layouts for each of the sub-modules and these layouts include various silkscreen and other documentation (or even segments of dimension). These additional data / layers should be brought into the higher level layouts when the modules are imported, instead of just the footprints and tracks, since they may have critical information (or just useful silkscreens) for the component's layout.

finger563:
Closes #397 

I've tested it locally and it does bring in the drawing layer:
CleanShot 2024-06-19 at 18 45 02@2x
from:
CleanShot 2024-06-19 at 18 46 01@2x

### DIFF
--- a/src/atopile/kicad_plugin/common.py
+++ b/src/atopile/kicad_plugin/common.py
@@ -77,6 +77,19 @@ def sync_track(
     source_board.Remove(new_track)
     return new_track
 
+# also pull in any silkscreen items
+def sync_drawing(
+    source_board: pcbnew.BOARD,
+    drawing: pcbnew.DRAWINGS,
+    target_board: pcbnew.BOARD
+) -> pcbnew.DRAWINGS:
+    """Sync a drawing to the target board."""
+    new_drawing: pcbnew.DRAWINGS = drawing.Duplicate().Cast()
+    new_drawing.SetParent(target_board)
+    new_drawing.SetLayer(drawing.GetLayer())
+    target_board.Add(new_drawing)
+    source_board.Remove(new_drawing)
+    return new_drawing
 
 #TODO: There must be a better way to update net of fill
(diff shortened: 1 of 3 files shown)


src/atopile/kicad_plugin/common.py
```diff
@@ -77,6 +77,19 @@ def sync_track(
     source_board.Remove(new_track)
     return new_track
 
+# also pull in any silkscreen items
+def sync_drawing(
```
mawildoer, line 81: Is this where you were thinking of using the `append` function? https://github.com/atopile/atopile/issues/397#issuecomment-2179397447

finger563: Yea I was just trying to figure out how id get everything to come in that I've added in the modules layout. This change will make it include drawings which is good, but as discussed in discord, any additional footprints may not be captured.

finger563: That was going to be an alternate implementation, but instead I think this works as intended.

src/atopile/kicad_plugin/pushgroup.py
```diff
@@ -63,9 +64,11 @@ def Run(self):
...
             )
 
-            for track in g.GetItems():
-                if isinstance(track, pcbnew.PCB_TRACK):
-                    sync_track(source_board, track, target_board)
+            for i in g.GetItems():
+                if isinstance(i, pcbnew.PCB_TRACK):
+                    sync_track(source_board, i, target_board)
+                elif isinstance(i, pcbnew.DRAWINGS):
+                    sync_drawing(source_board, i, target_board)
```
mawildoer, line 71: Do we still also need to ‘Move’ the drawings as well?

mawildoer (approved):

PR #458, opened 2024-08-05
add description if available from non-database components

napowderly:
Currently for local components, the comment field is blank in the BOM output, modifying to use 'description' if available.

### DIFF
--- a/src/atopile/components.py
+++ b/src/atopile/components.py
@@ -309,7 +309,10 @@ def get_user_facing_value(addr: AddrStr) -> str:
         else:
             return "?"
 
-    return get_specd_value(addr)
+    try:
+        return instance_methods.get_data(addr, "description")
+    except KeyError:
+        return "?"
 
 
 # FIXME: this function's requirements might cause a circular dependency
--- a/src/atopile/components.py
+++ b/src/atopile/components.py
@@ -306,13 +306,10 @@ def get_user_facing_value(addr: AddrStr) -> str:
         db_data = _get_generic_from_db(addr)
         if db_data:
             return db_data.get("description", "")
-        else:
-            return "?"
 
-    try:
-        return instance_methods.get_data(addr, "description")
-    except KeyError:
-        return "?"
+    instance = instance_methods.get_instance(addr)
+    doc_string = instance.supers[0].obj_def.doc_string
+    return doc_string.split("\n")[0] or "?"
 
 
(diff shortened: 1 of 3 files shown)


src/atopile/components.py
```diff
@@ -309,7 +309,10 @@ def get_user_facing_value(addr: AddrStr) -> str:
         else:
             return "?"
 
-    return get_specd_value(addr)
+    try:
+        return instance_methods.get_data(addr, "description")
```
mawildoer, line 313: Is this something you want a specific "description" variable for, or perhaps the doc-string instead?

sethfischer: Bear in mind the doc-string can be wordy. e.g.

```
module RpiHat:
    """
    Vehicle Management Computer Raspberry Pi HAT.

    [Vehicle management computer HAT datasheet](https://rover.fischer.nz/en/latest/reference/vmc-hat-datasheet.html).
    Conforming with the HAT+ specification dated 2023-12-20.

    # Reference

    * [Raspberry Pi HAT+ Specification](https://datasheets.raspberrypi.com/hat/hat-plus-specification.pdf)
    * [Raspberry Pi 40-pin header pin-out](https://github.com/raspberrypi/documentation/blob/c2ef9202f16e49e664ae604dff3bb4c3ec1f1fab/documentation/asciidoc/computers/os/using-gpio.adoc)
    * [Raspberry Pi 3B mechanical drawing](https://datasheets.raspberrypi.com/rpi3/raspberry-pi-3-b-mechanical-drawing.pdf)
    """
```

sethfischer: Parsing out the first line of the doc-string might be a good option.

mawildoer: I like that idea!

sethfischer: Ato is not Python but parsing out the first line is mentioned in  [PEP 257 – Docstring Conventions](https://peps.python.org/pep-0257/#multi-line-docstrings) if we're looking for a precedent.

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line. The summary line may be on the same line as the opening quotes or on the next line. The entire docstring is indented the same as the quotes at its first line (see example below).

PR #475, opened 2024-09-24
Use MPN instead of ? in BOM generation

LK:
This makes it easier on the BOM matching workflows of JLCPCB et al.


hexdae (approved):

mawildoer:
Thanks for the PR folks!

PR #482, opened 2024-10-03
nit: Add note in docs if issue wrt kicad plugin installation

computabeast:
Had to look through the code to figure out how to add the plugin to KiCad, added a quick note in the documentation if someone else has the same issue.


mawildoer:
Thanks a heap @computabeast ! 🚀 🙌

This flew under my radar for a little but I tweaked my notifications, so I make sure I get to it quicker in the future.

PR #492, opened 2024-10-24
Update URL for components API

sam-mellor:


napowderly (approved):
Tested on SpaceBMS, built successfully, did not test any deeper than that.

mawildoer (approved):

PR #565, opened 2024-11-21
Picker: Add mechanism to bypass picking container modules

sam-mellor:
Creates a new `skip_self_pick` trait, used for `MultiCapacitor`, which skips picking for the node itself, but not for any children. Fixes picking a single cap with capacitance equal to the sum of the child caps.

### DIFF
--- a/src/faebryk/libs/picker/picker.py
+++ b/src/faebryk/libs/picker/picker.py
@@ -148,11 +148,7 @@ class has_part_picked_remove(has_part_picked.impl()):
 
 
 class skip_self_pick(Module.TraitT.decless()):
-    """
-    Indicates that a node exists only to contain children, and shouldn't itself be picked
-    """
-
-    ...
+    """Indicates that a node exists only to contain children, and shouldn't itself be picked"""
 
 
 def pick_module_by_params(module: Module, options: Iterable[PickerOption]):


src/faebryk/libs/picker/picker.py
```diff
@@ -147,6 +147,17 @@ def mark_no_pick_needed(module: Module):
         )
 
 
+class is_virtual(Module.TraitT):
```
mawildoer, line 150: This should be a `decless` trait
```suggestion
class is_virtual(Module.TraitT.decless()):
```

mawildoer (approved):

iopapamanoglou:
Virtual is a very generic term for what is done here.
How about something like: `force_pick_children`

iopapamanoglou (changes requested):
See comment

iopapamanoglou:
`skip_self_pick` is fine

src/faebryk/libs/picker/picker.py
```diff
@@ -147,6 +147,14 @@ def mark_no_pick_needed(module: Module):
         )
 
 
+class skip_self_pick(Module.TraitT.decless()):
+    """
+    Indicates that a node exists only to contain children, and shouldn't itself be picked
+    """
+
+    ...
```
mawildoer, line 155: This isn't required. The doc string is sufficient

mawildoer (approved):

iopapamanoglou (approved):

PR #567, opened 2024-11-22
CLI: print non-ato group exception tracebacks on exit

mawildoer:
Screenshot 2024-11-21 at 16 00 39


mawildoer:
Potential future improvements I explored, but decided weren't worth it:
- Suppressing stack-sections between `errors.py` and `builtins`, which would reduce noise since they're infra
- Printing a tree of the exception group

PR #585, opened 2024-11-26
Better fp-lib-table management

sam-mellor:
Entries named `lcsc` in the `fp-lib-table` for each build config are now rewritten if they've drifted, and the entry points to the correct location (`lib/footprints/lcsc.pretty`, sibling to the PCB path).

### DIFF
--- a/src/faebryk/libs/app/pcb.py
+++ b/src/faebryk/libs/app/pcb.py
@@ -122,7 +122,7 @@ def include_footprints(pcb_path: Path):
         )
 
     # TODO make more generic, this is very lcsc specific
-    fppath = pcb_path.parent / "lib" / "footprints/lcsc.pretty"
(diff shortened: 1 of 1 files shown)


src/faebryk/libs/app/pcb.py
```diff
@@ -122,9 +122,7 @@ def include_footprints(pcb_path: Path):
         )
 
     # TODO make more generic, this is very lcsc specific
-    from faebryk.libs.picker.lcsc import LIB_FOLDER as LCSC_LIB_FOLDER
-
-    fppath = LCSC_LIB_FOLDER / "footprints/lcsc.pretty"
+    fppath = pcb_path.parent / "lib" / "footprints/lcsc.pretty"
```
mawildoer, line 125: Same as the other PR: no slashes in the RHS of paths

mawildoer (approved):

PR #587, opened 2024-11-26
Collect build paths in class

sam-mellor:
Collects build paths in a class. Paths are expected to be accessed via a BuildContext, but can be constructed separately for custom build mechanisms.

### DIFF
--- a/src/faebryk/libs/app/pcb.py
+++ b/src/faebryk/libs/app/pcb.py
@@ -5,12 +5,11 @@ import logging
 import os
 import sys
 from pathlib import Path
-from typing import Any, Callable
+from typing import TYPE_CHECKING, Any, Callable
 
 import psutil
 
 import faebryk.library._F as F
-from atopile.config import BuildPaths
 from faebryk.core.graph import Graph
 from faebryk.core.module import Module
 from faebryk.core.node import Node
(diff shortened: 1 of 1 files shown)


src/faebryk/libs/examples/buildutil.py
```diff
@@ -21,17 +22,25 @@
...
 lcsc.BUILD_FOLDER = BUILD_DIR
 lcsc.LIB_FOLDER = BUILD_DIR / Path("kicad/libs")
 lcsc.MODEL_PATH = None
 
+build_paths: BuildPaths = BuildPaths(
+    layout=KICAD_SRC / Path(f"{NAME}.kicad_pcb"),
+    lock_file=None,
+    build=BUILD_DIR,
+    output_base=BUILD_DIR / NAME,
+    netlist=BUILD_DIR / Path("faebryk/faebryk.net"),
```
mawildoer, line 38: The right side of the operator can just be a string, it doesn't need to be a `Path`
Additionally / more importantly, the strings on the RHS shouldn't contain slashes because it'll fuck w/ Windows compatibility

src/faebryk/libs/app/pcb.py
```diff
@@ -10,6 +10,7 @@
 import psutil
 
 import faebryk.library._F as F
+from atopile.config import BuildPaths
```
mawildoer, line 13: I'd probably recommend guarding this with a `if TYPING:`

mawildoer (approved):

Issue #547, opened 2024-11-19
Improve full-names of nodes


mawildoer:
- [ ] Remove `*.` prefix from the start of things
- [ ] Remove root node's name (It's the same for everything within a design)

iopapamanoglou:
> Remove root node's name (It's the same for everything within a design)
That's not true for non-bound objects

> Remove *. prefix from the start of things
That doesn't exist anymore right?
Thats now hex(id)[-4:] or something similar

iopapamanoglou:
I have an idea!

- Let's make a 'is_application_root' trait. 
- Add in the c++ Node a flag, for skip_from_full_name (workaround because C++ has no access to trait atm)
- make the trait set that flag on obj_set
- check in 'get_full_name' whether any node contains the skip and skip the respective nodes from the name

PR #592, opened 2024-11-27
Improve get_full_name to remove prefixed *HEX on nodes with `is_app_root` trait

mawildoer:
Fixes: https://github.com/atopile/atopile/issues/547

- Removes `atopile_name` PCB property in favor of `atopile_address`
- Replaces `.get_relative_address(app)`


src/faebryk/core/cpp/__init__.py
```diff
@@ -117,7 +117,9 @@ def compile_and_load():
                 is_pyi=True,
             )
         )
-        run_live(["ruff", "check", "--fix", pyi_out], logger=logger)
+        run_live(
```
mawildoer, line 120: This fixes a bug that was causing the formatter to fail on my machine. I think it was using the global python for some reason?

src/faebryk/exporters/netlist/graph.py
```diff
@@ -164,9 +162,7 @@ def name(self) -> str:
         Prefixes and suffixes are joined with a "-" if they exist.
         """
         return "-".join(
-            str(n)
-            for n in [self.prefix, self.base_name or "net", self.suffix]
-            if n is not None
+            str(n) for n in [self.prefix, self.base_name or "net", self.suffix] if n
```
mawildoer, line 165: This should test "truthiness", rather than is False, because we also don't want an empty string

src/faebryk/core/cpp/include/graph/graph.hpp
```diff
@@ -82,6 +82,7 @@ class Node {
   private:
     std::optional<nb::object> py_handle{};
     std::optional<Type> type{};
+    bool is_app_root = false;
```
iopapamanoglou, line 85: I would rename this to `no_include_parents_in_full_name`

iopapamanoglou (changes requested):
Congrats on the first cpp change! :)

just rename the flag (&setter&getter) within C++ and then its fine for now

mawildoer:
> Congrats on the first cpp change! :)
> 
> just rename the flag (&setter&getter) within C++ and then its fine for now

First of many I hope! 🚀

PR #593, opened 2024-11-28
Better exception handling

sam-mellor:
Integrates error handling between `atopile` and `faebryk` and improves handling of issues in user-provided code.

- `muffle_fatalities` and co are replaced by a custom logging handler, logging formatter, and exception hook
- errors arising from user app construction are re-raised as `UserPythonConstructionError`
- `UserPythonXError` are given special treatment for logging / traceback display
- logging presentation is extracted from faebryk to exist in `atopile.cli` only
- miscellaneous smaller improvements

Next:
- update all errors arising from user input to derive from `UserException`
- ensure all errors are raised rather than logged
- document a consistent taxonomy of error sources and their respective positions in the exception class hierarchy
- test outputs from more exception types

### DIFF
--- a/test/test_errors.py
+++ b/test/test_errors.py
@@ -55,7 +55,7 @@ def test_build_error_logging(build_name: str):
         [sys.executable, "-m", "atopile", "build", "-b", build_name],
         capture_output=True,
         text=True,
-        env=os.environ,
+        env={**os.environ, "ATO_NON_INTERACTIVE": "1"},
     )
 
     # single error


src/atopile/cli/build.py
```diff
@@ -94,21 +94,22 @@ def _init_python_app(build_ctx: BuildContext) -> Module:
...
             f"Build entry {build_ctx.entry.file_path} is not a module we can instantiate"
         )
 
-    app = app_class()
+    try:
+        app = app_class()
+    except Exception as e:
+        raise errors.UserPythonConstructionError(
+            f"Cannot construct build entry {build_ctx.entry.file_path}"
+        ) from e
```
mawildoer, line 112: Chaining exceptions seems to have variable mileage w/ debugging. I'm not entirely sure of a better approach for now, but there seems to be some ability to modify that with the exception attributes: https://docs.python.org/3/library/exceptions.html#exception-context

src/atopile/cli/logging.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+class AtoLogFormatter(logging.Formatter):
+    def __init__(self):
+        super().__init__(fmt="%(message)s", datefmt="[%X]")
+
+    def format(self, record: logging.LogRecord) -> str:
+        message = ""
+
+        if title := getattr(record, "title", None):
```
mawildoer, line 163: This isn't quite what I imagined, need to look further

src/atopile/errors.py
```diff
@@ -215,91 +166,33 @@ class UserPythonLoadError(UserException):
...
 
 _logged_exceptions: set[tuple[Type[Exception], tuple]] = set()
 
 
-def _log_user_errors(
-    ex: UserException | ExceptionGroup,
-    logger: logging.Logger,
-    de_dup: bool = True,
-):
+def _log_user_errors(ex: _BaseBaseUserException | ExceptionGroup, de_dup: bool = True):
```
mawildoer, line 184: Do we still need this? Shouldn't it be handled at the logging handler

src/atopile/errors.py
```diff
@@ -359,28 +256,24 @@ def muffle_fatalities():
         )
         do_exit = True
 
-    except Exception as ex:
+    except* Exception as ex:
```
mawildoer, line 259: This re-casts the underlying exception as a new ExceptionGroup, so the debugger only sinks to here
We should ideally not use this * syntax here

src/atopile/errors.py
```diff
@@ -392,6 +285,6 @@ def log_ato_errors():
     """
```
mawildoer, line 285: Maybe remove this

mawildoer (commented):
Check debugging behaviour of:
- UserException
- UserPythonException
- Exception

src/atopile/errors.py
```diff
@@ -215,91 +166,33 @@ class UserPythonLoadError(UserException):
...
 _logged_exceptions: set[tuple[Type[Exception], tuple]] = set()
 
 
-def _log_user_errors(
-    ex: UserException | ExceptionGroup,
-    logger: logging.Logger,
-    de_dup: bool = True,
-):
+def _log_user_errors(ex: _BaseBaseUserException | ExceptionGroup, de_dup: bool = True):
     """Helper function to consistently write errors to the log"""
```
mawildoer, line 185: Consider overriding exception handler

src/atopile/cli/build.py
```diff
@@ -94,21 +92,22 @@ def _init_python_app(build_ctx: BuildContext) -> Module:
         app_class = import_from_path(
             build_ctx.entry.file_path, build_ctx.entry.entry_section
         )
-    except (FileNotFoundError, ImportError) as e:
-        raise errors.UserPythonLoadError(
+    except Exception as e:
+        raise errors.UserPythonModuleError(
```
mawildoer, line 96: I'd prefer a split with an explicit error for things like `FileNotFound` error because it gives people something better to go chase down

src/atopile/cli/cli.py
```diff
@@ -1,31 +1,22 @@
-import logging
+import logging  # noqa: I001
```
mawildoer, line 1: Explain `noqa` in a comment

src/atopile/cli/excepthook.py
```diff
@@ -0,0 +1,103 @@
...
+from types import ModuleType
+
+import rich
+
+from atopile import telemetry
+from atopile.cli.logging import logger
+from atopile.errors import UserPythonModuleError, _BaseBaseUserException
+
+
+def in_debug_session() -> ModuleType | None:
```
mawildoer, line 12: I assume there's a good reason this isn't just `faebryk.exceptions...`, but tldr; because it's importing other things we don't want yet

I might advocate for a lazy import instead

src/atopile/cli/excepthook.py
```diff
@@ -0,0 +1,103 @@
...
+            telemetry.log_telemetry()
+
+        rich.print(
+            "\n\nUnfortunately errors ^^^ stopped the build. "
+            "If you need a hand jump on [#9656ce]Discord! https://discord.gg/mjtxARsr9V[/] :wave:"
+        )
+        sys.exit(1)
+
+
+sys.excepthook = handle_exception
```
mawildoer, line 103: Should we chain this to any existing `excepthook` instead of setting directly?

src/atopile/cli/excepthook.py
mawildoer, line 1: Comment for self/for cleanup later

This file feels generally very tangled with `faebryk.libs.exceptions`, maybe worth moving the infra side of things down there

src/atopile/cli/logging.py
```diff
@@ -0,0 +1,225 @@
...
+        r"(?P<Node>([/</>]))",
+        r"[?=\|](?P<Type>([a-zA-Z_0-9]+))[?=\>]",
+        r"[\.](?P<Child>([a-zA-Z_0-9]+))[?=\[]",
+        r"[\|](?P<Parent>([a-zA-Z_0-9]+))[?=\.]",
+        r"[?<=*.](?P<Root>(\*))",
+        r"[?=\[](?P<Number>([0-9]+))[?=\]]",
+    ]
+
+
+class AtoLogHandler(RichHandler):
```
mawildoer, line 44: nit: IMO this could be `LogHandler`, it's context is its module

test/test_errors.py
```diff
@@ -0,0 +1,96 @@
...
+from atopile.cli.build import _init_python_app
+from atopile.cli.common import create_build_contexts
+
+runner = CliRunner(mix_stderr=False)
+
+PROJECT_DIR = Path("test/common/resources/test-project")
+
+
+@pytest.fixture()
+def from_project_dir(monkeypatch: pytest.MonkeyPatch):
```
mawildoer, line 19: This looks like it might cause some pain at some point w/ xdist with conflicting simultaneous builds.

What about `tmp_file` instead, and copying the project to a unique tmp location for each fixture?

src/faebryk/libs/exceptions.py
```diff
@@ -13,33 +11,42 @@
...
+        Return a frozen version of this error.
+        """
+        return (self.__class__, self.message, self._title)
 
-        if debugpy.is_client_connected():
-            return debugpy
 
-    return None
+class UserResourceException(UserException):
+    """Indicates an issue with a user-facing resource, e.g. layout files."""
```
mawildoer, line 50: This is a very handy and, imo low-level util. It probably belongs here more than `errors`

src/atopile/front_end.py
```diff
@@ -871,10 +873,12 @@ def visitAssign_stmt(self, ctx: ap.Assign_stmtContext) -> KeyOptMap:
         elif assignable_ctx.string() or assignable_ctx.boolean_():
             # Check if it's a property or attribute that can be set
             if not try_set_attr(target, assigned_name, value):
-                errors.UserException.from_ctx(
-                    ctx,
-                    f"Ignoring assignment of {value} to {assigned_name} on {target}",
-                ).log(to_level=logging.WARNING)
+                logger.warning(
```
mawildoer, line 876: `downgrade` instead?

mawildoer (approved):

PR #602, opened 2024-11-30
Create PCB file on build if it doesn't exist

sam-mellor:
Creates a minimal PCB file, and emits a warning, if an existing layout file isn't found.

### DIFF
--- a/src/faebryk/libs/kicad/fileformats.py
+++ b/src/faebryk/libs/kicad/fileformats.py
@@ -1162,7 +1162,7 @@ class C_kicad_pcb_file(SEXP_File):
     kicad_pcb: C_kicad_pcb
 
     @staticmethod
-    def create_skeleton(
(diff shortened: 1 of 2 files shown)


src/faebryk/exporters/bom/jlcpcb.py
mawildoer, line 1: Seems there's a little extra here from the other PR

src/faebryk/libs/kicad/fileformats.py
```diff
@@ -1159,6 +1161,20 @@ class C_group:
 
     kicad_pcb: C_kicad_pcb
 
+    @staticmethod
+    def from_generator(
```
mawildoer, line 1165: I named a similar function `skeleton` in the schematic fileformats

mawildoer (approved):

PR #616, opened 2024-12-03
Packaging: Complete switch to `uv`

sam-mellor:
Switches remaining workflows to use `uv` for building/installing. Adds uv lockfile to repo.

Note `uv.lock` isn't yet supported by dependabot, so we'll lose automatic dependency updating absent implementing some workaround.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,11 +109,18 @@ Find our [documentation](https://atopile.io/), [installation video](https://www.
 
 `atopile` is on pypi.org: https://pypi.org/project/atopile/
 
-### Most Basic Installation
+### Installation
 
-`atopile` requires *python3.12* or later, which you can install using your package manager or from [python.org](https://www.python.org/downloads/).
+`atopile` is published to [pypi.org](https://pypi.org/project/atopile/). We recommend installing into an isolated environment, e.g. with `uv`:
 
-Then just `pipx install atopile` and you're good to go!
+```sh
+uv tool install atopile
+```
+
+Or with `pipx` (requires Python 3.12 or later):
+```sh
+pipx install atopile
+```
 
 ## ❓ Why Atopile?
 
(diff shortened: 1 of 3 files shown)


.github/workflows/wheels.yml
```diff
@@ -29,11 +29,18 @@ jobs:
...
+    - name: Install uv
+      uses: astral-sh/setup-uv@v4
+      with:
+        version: '0.5.5'
+        enable-cache: true
+        cache-dependency-glob: "uv.lock"
+
     - name: Build SDist
-      run: pipx run build --sdist
+      run: uv build --sdist
```
mawildoer, line 40: Weird we're doing this in two steps:
1. Does the `sdist` work? Do we supply binaries of the cpp content? etc...
2. Do we need `sdist` at all

dockerfiles/Dockerfile.kicad
```diff
@@ -17,29 +20,20 @@ USER root
...
 git config --global --add safe.directory /github/workspace
 EORUN
 
-RUN <<EOF
-curl -LsSf https://astral.sh/uv/install.sh | sh
-uv venv --python 3.12 /venv
-EOF
+COPY --from=ghcr.io/astral-sh/uv:0.5.5 /uv /bin/uv
+
+RUN uv python install 3.12
```
mawildoer, line 33: Possible to get this from the `pyproject.toml` too?

pyproject.toml
```diff
@@ -119,26 +119,9 @@ docs = ["mkdocs", "mkdocs-material", "mkdocs-drawio-file >= 1.5.2"]
 ato = "atopile.cli.cli:main"
 faebryk = "faebryk.tools.main:__main__"
 
-[tool.hatch]
```
mawildoer, line 122: nit: I think we've maintained a chapter-like styling through this PR, which this conforms with

mawildoer (commented):
- [ ] Docs updates to use `uv`

mawildoer (approved):

PR #617, opened 2024-12-04
CI: Update workflow triggers

sam-mellor:
Runs `Continuous Delivery` workflow on PRs (minus publishing of built package), and all workflows when the workflow definition changes. This reduces the circumstances where CI checks can be invisibly broken in a PR.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -8,6 +8,8 @@ on:
      - main
     tags:
       - 'v*'
+    paths:
+      - '.github/workflows/cd.yml'
 
 jobs:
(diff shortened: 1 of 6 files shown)


.github/workflows/cd.yml
```diff
@@ -1,25 +1,18 @@
 name: Continuous Delivery
 
 on:
+  workflow_dispatch:
+  pull_request:
   push:
     branches:
-      - 'main'
-    tags:
```
mawildoer, line 7: Let's put this back

mawildoer (approved):

PR #645, opened 2024-12-06
rm and re-create fresh uv.lock

mawildoer:


uv.lock
```diff
@@ -581,25 +582,25 @@ wheels = [
 
 [[package]]
 name = "fake-useragent"
-version = "1.5.1"
+version = "2.0.0"
```
mawildoer, line 585: @sam-mellor @iopapamanoglou does this need more scrutiny?

iopapamanoglou (approved):

uv.lock
iopapamanoglou: opened #648

PR #647, opened 2024-12-06
Make layout mapping case-insensitive

mawildoer:
I don't think we need to handle case sensitivity yet, but it could be problem in the future

I didn't see a good way to do this with `pathlib`

### DIFF
--- a/src/faebryk/libs/util.py
+++ b/src/faebryk/libs/util.py
@@ -1523,13 +1523,17 @@ def get_module_from_path(
     """
     Return a module based on a file path if already imported, or return None.
     """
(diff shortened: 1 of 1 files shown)


src/faebryk/libs/util.py
```diff
@@ -1527,7 +1527,8 @@ def get_module_from_path(
     try:
         module = find(
             sys.modules.values(),
-            lambda m: getattr(m, "__file__", None) == sanitized_file_path,
+            lambda m: str(getattr(m, "__file__", None)).lower()
+            == sanitized_file_path.lower(),
```
iopapamanoglou, line 1531: Pathlib.samefile

iopapamanoglou (approved):

PR #654, opened 2024-12-09
Accept unknown keys by default in layout files

mawildoer:
Hopefully improves file stability


src/faebryk/libs/kicad/fileformats.py
```diff
@@ -705,13 +710,15 @@ class C_rotate:
 
 @dataclass
 class C_kicad_pcb_file(SEXP_File):
+    @dataclass_json(undefined=Undefined.INCLUDE)
```
iopapamanoglou, line 713: Am I understanding correctly that we are abusing the dataclass_json decorator everywhere for the catch all?

src/faebryk/libs/kicad/fileformats.py
```diff
@@ -660,6 +660,7 @@ class C_gr_poly(C_polygon):
         options: Optional[C_options] = None
         primitives: Optional[C_gr] = None
         # TODO: primitives: add: gr_line, gr_arc, gr_circle, gr_rect, gr_curve, gr_bbox
+        unknown: CatchAll = None
```
mawildoer, line 663: We're abusing this catch-all type from `dataclass_json`. Can make our own if we care. It's just a type (or really a pair)

src/faebryk/libs/sexp/dataclass_sexp.py
```diff
@@ -107,6 +109,7 @@ def _convert(
                 and len(args) == 2
```
mawildoer, line 109: I was tempted to flatten this function into a single loop, since it's pretty complicated rn to follow, but started on it and realised there were enough corner cases that now wasn't the time to touch it

src/faebryk/libs/sexp/dataclass_sexp.py
```diff
@@ -275,31 +281,36 @@ def _decode[T](
...
-        origin = get_origin(f.type)
-        if origin is list:
-            vs = []
-            next_val = v
-            # consume all values
-            while next_val is not None:
-                vs.append(next_val)
-                next_val = it.next(1, None)
-            out = _convert(vs, f.type, stack, f.name, sp)
+        if (
```
mawildoer, line 289: This is a simple `continue` -> `else` replacement so there's only one loop exit point

src/faebryk/libs/sexp/dataclass_sexp.py
```diff
@@ -275,31 +281,36 @@ def _decode[T](
                 value_dict[name] = out
 
     # Positional
-    for f, v in (it := zip_non_locked(positional_fields.values(), pos_values.values())):
+    for f, (sexp_i, v) in (
+        it := zip_non_locked(positional_fields.values(), pos_values.items())
```
mawildoer, line 285: Note, the second element of `it` is now a tuple of the sexp index and value

src/faebryk/libs/sexp/dataclass_sexp.py
```diff
@@ -275,31 +281,36 @@ def _decode[T](
...
+            if it.next(0) is None:
+                raise ValueError(f"Unexpected symbol {v}")
         else:
-            out = _convert(v, f.type, stack, f.name, sp)
+            # positional list = var args
+            origin = get_origin(f.type)
+            if origin is list:
+                vs = [v]
+                # consume all values
+                while (next_val := it.next(1, None)) is not None:
```
mawildoer, line 306: Update this loop to get just the values from `it[2]`

src/faebryk/libs/sexp/dataclass_sexp.py
```diff
@@ -311,13 +322,18 @@ def _decode[T](
                 f" {sp.assert_value} but is {value_dict[f.name]}"
             )
 
+    # Unprocessed values should be None is there aren't any,
+    # so they don't get reproduced etc...
+    if catch_all_fields and not unprocessed:
```
mawildoer, line 327: Convert an empty catch-all to `None`, so it doesn't show up in outputs

iopapamanoglou (approved):

PR #671, opened 2024-12-11
Core: Improve pick speed

iopapamanoglou:
# Core: Improve pick speed

- Better cache for superset
- Disable slow printing

Extras:
- Fix LDO enable
- Better perf printing
- new perf test
- Sugar for decouple
- Better contradiction error
- FIXME: Disabled rated_voltage decouple cap


src/faebryk/library/ElectricPower.py
```diff
@@ -25,6 +33,8 @@ def decouple(self):
         def is_implemented(self):
             return not self.obj.has_trait(F.is_decoupled)
 
+        # TODO implement merging
```
mawildoer, line 36: This looks like it might be important enough to not forget about

src/faebryk/library/RP2040_ReferenceDesign.py
```diff
@@ -119,39 +108,16 @@ def __preinit__(self):
...
-        multi_cap = F.MultiCapacitor(2)
-        self.rp2040.power_core.decoupled.decouple().specialize(
-            multi_cap
-        ).set_equal_capacitance_each(range_100nF)
-        caps_100nF.extend(multi_cap.capacitors)
-        self.rp2040.core_regulator.power_out.decoupled.decouple().capacitance.constrain_subset(
-            L.Range.from_center_rel(1 * P.uF, 0.05)
-        )
-
+        self.rp2040.power_io.decoupled.decouple(6).explicit(100 * P.nF)
```
mawildoer, line 114: These MultiCap changes seem unrelated, and I think warrant more scrutiny, but we won't do it now because it's entangled with the performance updates. Let's get back to it later

test/core/test_performance.py
```diff
@@ -276,3 +280,60 @@ def test_no_connect(self):
         timings.times["total"] = sum(all_times)
 
         logger.info(f"\n{timings}")
+
+
+# DONT COMMIT
```
mawildoer, line 285: Don't commit? Perhaps instead mark slow or use an example instead?

mawildoer (commented):
Only one that matters is the test w/ "DONT COMMIT" written above it

mawildoer (approved):

PR #677, opened 2024-12-12
Picker: Send `None` for unconstrained parameters

sam-mellor:
Sends `None` in lieu of serialized `P_Set` to the API when a parameter is unconstrained. This allows the backend to bypass filtering when any parameter value is permitted.

### DIFF
--- a/src/faebryk/libs/picker/api/picker_lib.py
+++ b/src/faebryk/libs/picker/api/picker_lib.py
@@ -150,7 +150,10 @@ def _find_component_by_params[T: BaseParams](
         raise PickError(f"Module is not a {cmp_class.__name__}", cmp)
 
     fps = get_package_candidates(cmp)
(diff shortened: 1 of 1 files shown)


src/faebryk/core/parameter.py
```diff
@@ -1508,6 +1520,10 @@ def int_to_subscript(i: int) -> str:
     def domain_set(self) -> P_Set:
         return self.domain.unbounded(self)
 
+    def to_set(self, solver: "Solver") -> P_Set | None:
+        as_literal = solver.inspect_get_known_supersets(self, force_update=False)
+        return None if as_literal == self.domain_set() else as_literal
+
```
iopapamanoglou, line 1526: I don't like this being here.
But we can leave it here for now.
Rename it to: `get_last_known_deduced_superset`

iopapamanoglou (changes requested):

iopapamanoglou (approved):

PR #683, opened 2024-12-13
Picker: non-recursive, solver more predicate deduction

iopapamanoglou:
- More deducible predicates for unconstrained parameters
- Picker is now using pick trees, ready for heuristics
- Only add pickers to modules that have a chance of being picked


src/faebryk/libs/picker/api/common.py
```diff
@@ -271,9 +271,7 @@ def check_compatible_parameters(
 
     # check for every param whether the candidate component's range is
     # compatible by querying the solver
-    anded = And(
-        *(m_param.operation_is_superset(c_range) for m_param, c_range in param_mapping)
```
iopapamanoglou, line 275: if we pick a cap for X.capacitance, nobody is allowed to constrain_superset with a wider set than the part has so superset is not correct.

src/faebryk/libs/picker/api/pickers.py
```diff
@@ -43,15 +47,23 @@ def add_api_pickers(module: Module, base_prio: int = 0) -> None:
     # TODO: might want to remove this 'if' since the trait might be defined later
     # but for now handy to guard, since it makes logging more concise
     if module.has_trait(F.has_descriptive_properties):
```
iopapamanoglou, line 49: now only adds picker if if has the props, also doesnt add db picker for explicit stuff

src/faebryk/libs/picker/api/pickers.py
```diff
@@ -43,15 +47,23 @@ def add_api_pickers(module: Module, base_prio: int = 0) -> None:
...
+        props = module.get_trait(F.has_descriptive_properties).get_properties()
+        if "LCSC" in props:
+            logger.debug(f"Adding LCSC picker for {module.get_full_name()}")
+            module.add(
+                F.has_multi_picker(
+                    base_prio, ApiPicker(picker_lib.find_and_attach_by_lcsc_id)
+                )
+            )
+            return
+        if DescriptiveProperties.partno in props:
```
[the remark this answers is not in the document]
iopapamanoglou: Not sufficient, but seems to me when you provided a partno you kinda expect a pick by it.
So when the picker fails it gives the appropriate error.
Having it be autopicked would be weird, so thats why only this is checked.

PR #687, opened 2024-12-14
Make `ato` projects build again - Part 1.

mawildoer:
Testing against: https://github.com/atopile/spin-servo-drive/pull/40

Fails, but gets soooo much further, and many intermediate designs should be functional now


src/atopile/front_end.py
```diff
@@ -323,13 +331,15 @@ def visitImport_stmt(
...
+            return KeyOptMap(
+                KeyOptItem.from_kv(li.ref, (li, ctx)) for li in lazy_imports
+            )
 
         else:
             # Standard library imports are special, and don't require a from path
             imports = []
             for collector, name_or_attr in iter_through_errors(ctx.name_or_attr()):
-                with collector():
+                with collector(), log_user_errors():
```
mawildoer, line 342: This change means that we end up with the errors + warnings properly in sequence and interlaced w/ any other build info

src/atopile/front_end.py
```diff
@@ -224,6 +224,10 @@ class NOTHING:
     """A sentinel object to represent a "nothing" return value."""
 
 
+class SkipPriorFailedException(Exception):
```
mawildoer, line 227: This exception helps us cut off execution of code-paths which fail because they rely on bit that have already failed

src/atopile/front_end.py
```diff
@@ -382,11 +392,22 @@ def survey(
...
-        for ref, item in surveyor.visit(ctx):
+        for ref, (item, item_ctx) in surveyor.visit(ctx):
             if ref in context.refs:
-                raise errors.UserKeyError.from_ctx(
-                    ctx, f"Duplicate declaration of {ref}"
-                )
+                ex = errors.UserKeyError.from_ctx(item_ctx, f'"{ref}" already declared')
+                # Downgrade the error if we're just importing
+                # the same thing multiple times
+                first_of_dup = context.refs[ref]
```
mawildoer, line 400: There's a few cases where, in the std library we have duplicated imports etc...
Ignoring that for now

src/atopile/front_end.py
```diff
@@ -432,33 +468,32 @@ class Bob(BasicsMixin, PhysicalValuesMixin, SequenceMixin, AtoParserVisitor):  #
 
     def __init__(self) -> None:
         super().__init__()
-        self._scopes = FuncDict[ParserRuleContext, Context]()  # type: ignore
-        self._python_classes = FuncDict[ap.BlockdefContext, Type[Component]]()  # type: ignore
-        self._node_stack = StackList[L.Node]()  # type: ignore
-        self._promised_params = FuncDict[L.Node, list[ParserRuleContext]]()  # type: ignore
```
mawildoer, line 438: 3.13 upgrade let me remove all these

src/atopile/front_end.py
```diff
@@ -490,16 +525,33 @@ def _build(self, context: Context, ref: Ref) -> L.Node:
...
+            errors._BaseBaseUserException, SkipPriorFailedException
+        ) as ex_acc:
             for param, (value, ctx) in self._param_assignments.items():
                 with ex_acc.collect(), ato_error_converter(), log_user_errors(logger):
                     if value is None:
-                        raise errors.UserKeyError.from_ctx(
+                        ex = errors.UserKeyError.from_ctx(
                             ctx, f"Parameter {param} never assigned"
                         )
+                        if param in self._promised_params:
```
mawildoer, line 545: Again, std lib issue. Don't complain unless we actually try use an undefined param

src/atopile/front_end.py
```diff
@@ -1011,7 +1086,7 @@ def visitConnectable(self, ctx: ap.ConnectableContext) -> L.ModuleInterface:
             return node
         elif numerical_ctx := ctx.numerical_pin_ref():
             pin_name = numerical_ctx.getText()
-            ref = Ref.from_one(pin_name)
+            ref = Ref(pin_name.split("."))
```
mawildoer, line 1089: This fixes dot accessing pins with `component.1`

src/atopile/front_end.py
```diff
@@ -1049,43 +1124,36 @@ def visitAssert_stmt(self, ctx: ap.Assert_stmtContext) -> KeyOptMap:
...
+        op_strs = [
+            cop.getChild(0).getChild(0).getText() for cop in ctx.compare_op_pair()
+        ]
+
+        predicates = []
+        for (lh, rh), op_str in zip(itertools.pairwise(exprs), op_strs):
+            match op_str:
                 case "<":
-                    op = operator.lt
+                    op = fab_param.LessThan
```
mawildoer, line 1140: IMO this is cleaner

src/atopile/front_end.py
```diff
@@ -1049,43 +1124,36 @@ def visitAssert_stmt(self, ctx: ap.Assert_stmtContext) -> KeyOptMap:
...
-        for lh_ctx, rh_ctx in itertools.pairwise(
-            itertools.chain([ctx.arithmetic_expression()], ctx.compare_op_pair())
-        ):
-            lh = _visit(lh_ctx.arithmetic_expression())
-            rh = _visit(rh_ctx.arithmetic_expression())
-
-            # HACK: this is a cheap way to get the operator string
-            operator_str: str = rh_ctx.getChild(0).getText()
-            match operator_str:
+        exprs = [
```
mawildoer, line 1127: Assertions were failing because the `lh` and `rh` sides were variable. This approach guarantees operands + an operator better

src/atopile/front_end.py
```diff
@@ -1049,43 +1124,36 @@ def visitAssert_stmt(self, ctx: ap.Assert_stmtContext) -> KeyOptMap:
...
-                    op = operator.contains
+                    op = fab_param.IsSubset
                 case _:
                     # We shouldn't be able to get here with parseable input
-                    raise ValueError(f"Unhandled operator {operator_str}")
+                    raise ValueError(f"Unhandled operator {op_str}")
 
             # TODO: should we be reducing here to a series of ANDs?
-            params.append(op(lh, rh))
+            predicates.append(op(lh, rh))
```
mawildoer, line 1154: I believe this is a more accurate term, @iopapamanoglou

src/atopile/front_end.py
```diff
@@ -1127,8 +1195,7 @@ def visitTerm(self, ctx: ap.TermContext) -> fab_param.ParameterOperatable:
 
     def visitPower(self, ctx: ap.PowerContext) -> fab_param.ParameterOperatable:
         if ctx.POWER():
-            base = self.visitFunctional(ctx.functional())
-            exp = self.visitFunctional(ctx.functional())
+            base, exp = map(self.visitFunctional, ctx.functional())
```
mawildoer, line 1198: Not just formatting, also fixes an issue w/ getting the right `functional` ctx

src/atopile/front_end.py
```diff
@@ -1137,8 +1204,38 @@ def visitFunctional(
         self, ctx: ap.FunctionalContext
     ) -> fab_param.ParameterOperatable:
         if ctx.name():
-            # TODO: implement min/max
-            raise NotImplementedError
+            name = self.visitName(ctx.name())
+            operands = [self.visitBound(b) for b in ctx.bound()]
+            if name == "min":
```
mawildoer, line 1209: We can probably come up with a better pattern for these functions. These are the only cases so far, but I could imagine expansion as we decide to expand the language

src/atopile/front_end.py
```diff
@@ -1197,18 +1294,29 @@ def visitDeclaration_stmt(self, ctx: ap.Declaration_stmtContext) -> KeyOptMap:
             )
 
         assigned_name = assigned_value_ref[0]
-        unit = self._get_unit_from_ctx(ctx.type_info().name_or_attr())
+        unit_ctx: ap.Name_or_attrContext = ctx.type_info().name_or_attr()
+        # TODO: @v0.4.0: remove this shim
+        unit_ref = self.visitName_or_attr(unit_ctx)
```
mawildoer, line 1299: I dislike the declaration syntax atm, which is why I've marked this as deprecated

Should consider how we can expand to domains etc... too

src/faebryk/core/node.py
```diff
@@ -552,7 +551,6 @@ def builder(self, op: Callable[[Self], Any]) -> Self:
 
     # printing -------------------------------------------------------------------------
 
-    @try_avoid_endless_recursion
```
mawildoer, line 555: This is implemented in `cpp` now, which means no more recursion risk

PR #700, opened 2024-12-17
Regression test packages and projects on main or with regression-required label

mawildoer:
# Got some wins on the board. Time to build from here

Screenshot 2024-12-16 at 16 41 49

## What is done:
1. Add regression mark in pytest (to ignore in regular pre-merge testing)
2. Add new regression workflow, which runs on main and on labeled PRs
4. Regression test a bunch of our repos


.github/workflows/pytest.yml
```diff
@@ -1,59 +1,55 @@
...
+        python-version-file: "pyproject.toml"
+
+    - name: Setup cmake
+      uses: jwlawson/actions-setup-cmake@v2
+
+    - name: Run pytest
+      id: pytest
+      continue-on-error: true
+      run: |
+        uv run --dev --no-editable --frozen pytest -m "not not_in_ci and not slow and not regression" --timeout=300
```
mawildoer, line 36: Primarily, this is just a dedent / format. This line is the only one which matters

++ "not regression"
++ "timeout"

.github/workflows/regression.yml
mawildoer, line 1: I'd love this not to be a copy of pytest. Wonder if we can reuse things?

.github/workflows/regression.yml
```diff
@@ -0,0 +1,67 @@
...
+on:
+  push:
+    branches: [main]  # Adjust this if your default branch has a different name
+  pull_request:
+    types: [labeled]
+  workflow_dispatch:  # Add manual trigger capability
+
+jobs:
+  regression:
+    runs-on: blacksmith-4vcpu-ubuntu-2204
```
mawildoer, line 12: Went fairly small on the CPUs here because I think it's likely we have one or two jobs taking waayyyy longer than the others

.github/workflows/regression.yml
```diff
@@ -0,0 +1,67 @@
...
+    steps:
+    # Untrusted action -> specific githash
+    - uses: snnaplab/get-labels-action@f426df40304808ace3b5282d4f036515f7609576
```
mawildoer, line 15: Maybe we should make this an action ourselves. Perhaps good publicity, not too complicated and fuck it's a PITA right now

.github/workflows/regression.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    - name: Setup cmake
+      uses: jwlawson/actions-setup-cmake@v2
+
+    - name: Run pytest
+      id: pytest
+      continue-on-error: true
+      # Regular tests are run as part of the pytest workflow
+      run: |
+        uv run --dev --no-editable --frozen pytest -m "not not_in_ci and regression" --timeout=900
```
mawildoer, line 48: Only other difference to `pytest` workflow

mawildoer:
Bypassed because the backend deployment was causing the failures in pytest action

PR #701, opened 2024-12-17
Tools: Better profile script

iopapamanoglou:
# Tools: Better profile script

Translate to python


mawildoer (approved):
- Looks like it's especially linuxy
- If it helps you, sweet!

PR #706, opened 2024-12-17
Library: Add explicit kicad part

iopapamanoglou:
I don't like this:

- needs global path
    - doesn't work in CI
    - weird global vars
- what is the expected behavior/usecase here?
    - because this will prohibit picking, this results in it being missing in the bom?


iopapamanoglou:
Needs rebase

PR #707, opened 2024-12-17
Fix: Net collision by instable lazy mifs

iopapamanoglou:
# Fix: Net collision by instable lazy mifs

Lazy mifs result in sometimes missing a connection to a net.
This new `add_or_get_nets` is more robust against that but still correct.


src/faebryk/exporters/netlist/graph.py
```diff
@@ -104,19 +104,32 @@ def get_kicad_obj(self):
...
+    to_check = set(interfaces)
+    bus_reprs = {}
+    while to_check:
+        interface = to_check.pop()
+        ifs = interface.get_connected()
+        bus_reprs[interface] = ifs
+        to_check.difference_update(ifs.keys())
+    nets_out = set()
+
+    for _, connected_mifs in bus_reprs.items():
```
mawildoer, line 117: nit; `.values()` - not worth updating for now

mawildoer (approved):

iopapamanoglou:
Does not work

PR #711, opened 2024-12-18
Make ato build again - Part 3 the Last Crusade

mawildoer:


src/atopile/_shim.py
```diff
@@ -47,7 +41,7 @@ def _is_int(name: str) -> bool:
     return True
 
 
-class _has_kicad_footprint_name_defined(F.has_footprint_impl):
+class has_local_kicad_footprint_named_defined(F.has_footprint_impl):
```
mawildoer, line 44: Hopefully we can replace this from this shim momentarily w/ your new explicit part pick

src/atopile/_shim.py
```diff
@@ -238,7 +249,7 @@ def value(self) -> L.Range:
 
     @value.setter
     def value(self, value: L.Range):
-        _alias_is(self.capacitance, value)
```
mawildoer, line 241: Better models existing ato behaviour

src/atopile/buildutil.py
```diff
@@ -21,6 +25,7 @@
     export_pick_and_place,
     export_step,
 )
+from faebryk.exporters.pcb.kicad.pcb import _get_footprint
```
mawildoer, line 28: Used as a temp measure to check footprint quality early and avoid late (eg. long) expensive failures

src/faebryk/exporters/pcb/kicad/pcb.py
```diff
@@ -319,83 +333,89 @@ def get_property_uuid(
...
-
-            at = C_xyr(x=x, y=y, r=0)
-
-            # separate components vertically within group
-            y += VERTICAL_SPACING
-
-            if comp_name in comps_changed:
-                # TODO also need to do geo rotations and stuff
-                at = comps_changed[comp_name].at
+        for err_collector, comp_name in iter_through_errors(
```
mawildoer, line 336: Dunno why the diff aligns so poorly. This should all just be indented one to add the error collection context. Double checking now

src/faebryk/libs/kicad/fileformats.py
```diff
@@ -1200,7 +1200,7 @@ class C_kicad_footprint_file(SEXP_File):
     class C_footprint_in_file(C_footprint):
         descr: Optional[str] = None
         tags: Optional[list[str]] = None
-        version: int = field(**sexp_field(assert_value=20240108), default=20240108)
```
mawildoer, line 1203: Shit, should re-add this. Will fix

src/faebryk/libs/sexp/dataclass_sexp.py
```diff
@@ -99,24 +100,36 @@ def _convert(
         if (origin := get_origin(t)) is not None:
             args = get_args(t)
             if origin is list:
-                return [_convert(_val, args[0], substack) for _val in val]
+                return [
```
mawildoer, line 103: This is as a backstop to TRY load stuff. I think we could take a more delicate approach in the future, but works well for now. See https://github.com/atopile/atopile/pull/711/files#diff-02f2b52a27305bdc76f47e7e4e5e3040a17a3b9b12bcb4cb3593b245e028e6a9

src/faebryk/exporters/pcb/kicad/pcb.py
mawildoer: Yep, confirmed there are no other changes

iopapamanoglou (approved):

PR #717, opened 2024-12-18
Add inductors to shims, and use L.Module as base component for all `ato` components and modules

mawildoer:


src/atopile/_shim.py
```diff
@@ -87,72 +87,37 @@ def handle_duplicate(
...
 
-    @L.rt_field
-    def attach_to_footprint(self):
-        return F.can_attach_to_footprint_via_pinmap(self.pinmap)
-
-    @L.rt_field
-    def has_designator_prefix(self):
-        return F.has_designator_prefix_defined(F.has_designator_prefix.Prefix.U)
+    def on_obj_set(self):
+        self.module = cast_assert(L.Module, self.obj)
```
iopapamanoglou, line 96: there is shortcut, self.get_obj(L.Module)

iopapamanoglou (approved):

src/atopile/_shim.py
mawildoer: Thanks! Will update in next iteration

PR #726, opened 2024-12-20
Library: Total max power consumption on bus

iopapamanoglou:
- Add `bus_max_current_consumption_sum`
- Add `F.is_dynamic_by_connections_sum`
- Add test


iopapamanoglou:
Works, solver is just not smart enough for it

mawildoer:
It seems like this could be made generically pretty easily, if we accepted moving the bus-oriented params into the trait (which gut feel makes a lot of sense to me)

I would be willing to introduce an invariant that `mifs` are node-oriented, and that anything that happens to be aliased-is across the nodes only happens to be so (eg. voltage is the same at all nodes, but it's still not looking at the bus)
This seems consistent to me with how decoupling works, for example and feels the most intutative

In that case, we could make a trait with the signature `is_dynamic_by_connection(op: Operation)`
We could potentially add a key (eg, keying by name is insufficent)
Then, within the trait we'd offer a public collection/accumulation/bus/cum param, which is the outcome of all those operations. For the `alias_is`, nothing happens - they're all just aliased to one another, so you'd never look
For the `Sum` or the likes, you can examine that param to get the bus current

Seems simpler, but I'm also certain you've considered some corner cases I'm missing

src/faebryk/libs/app/parameters.py
```diff
@@ -16,10 +16,15 @@
 
 logger = logging.getLogger(__name__)
 
+MIF_TRAIT = F.is_dynamic_by_connections_alias | F.is_dynamic_by_connections_sum
+
 
 def resolve_dynamic_parameters(graph: Graph):
     other_dynamic_params, connection_dynamic_params = partition(
-        lambda param_trait: isinstance(param_trait[1], F.is_dynamic_by_connections),
```
mawildoer, line 22: What about a hierarchy like

```python
class is_dynamic(Trait): ...

class is_dynamic_by_connection(is_dynamic): ...
# cut here and use `decless` if we made the generic version

class is_dynamic_by_connections_alias(is_dynamic_by_connection.impl()): ...

class is_dynamic_by_connections_sum(is_dynamic_by_connection.impl()): ...
```

The `isinstance` here is a bit of a smell, and very traitorous

PR #747, opened 2024-12-23
CLI: Improve `ato build` output

sam-mellor:
- quotes component representations in log output, to fix inconsistent highlighting
- quotes build targets separately
- fixes complaints about ambiguous module paths and missing entry section for `generics` when generating module map

### DIFF
--- a/src/faebryk/libs/exceptions.py
+++ b/src/faebryk/libs/exceptions.py
@@ -234,7 +234,7 @@ class downgrade[T: Exception](Pacman):
             exceptions = [exc]
 
         for e in exceptions:
-            self.logger.log(self.to_level, e, exc_info=exc)
+            self.logger.log(self.to_level, str(e), exc_info=e)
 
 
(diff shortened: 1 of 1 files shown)


src/atopile/buildutil.py
```diff
@@ -306,10 +309,10 @@ def consolidate_footprints(build_ctx: BuildContext, app: Module) -> None:
     elif lib_in_fptable and not lib_prefix_on_ids:
         # We could probably just remove the lib entry
         # but let's be conservative for now
-        logging.info(
+        logger.info(
             "It seems like this project is using a legacy footprint consolidation "
             "unnecessarily. You can likely remove the 'lib' entry from the "
-            "fp-lib-table file."
+            "'fp-lib-table' file."
```
mawildoer, line 315: nit; typically, we'd flip these and have double-inverted commas

mawildoer (approved):

PR #750, opened 2024-12-23
Don't run regression testing by default

mawildoer:
This currently sucks, because without the changes commented out in `settings.shared.json`, it doesn't collect marked-out-tests, and with them it, it won't run them

This should fix it:
https://github.com/microsoft/vscode-python/issues/21845


mawildoer:
Regression seems to all be working now

PR #758, opened 2024-12-24
Fix: Frontend: Pick caps

iopapamanoglou:
- Fixes footprint candidate shim
- Fixes picking trait

### DIFF
--- a/src/faebryk/libs/picker/picker.py
+++ b/src/faebryk/libs/picker/picker.py
@@ -27,6 +27,7 @@ from faebryk.core.parameter import (
 from faebryk.core.solver.solver import LOG_PICK_SOLVE, Solver
 from faebryk.libs.util import (
     ConfigFlag,
+    KeyErrorAmbiguous,
     KeyErrorNotFound,
     Tree,
     indented_container,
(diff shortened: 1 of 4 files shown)


src/faebryk/libs/picker/picker.py
```diff
@@ -268,9 +269,13 @@ def check_missing_picks(module: Module):
         )
         # no parent with picker
         and not try_or(
-            lambda: m.get_parent_with_trait(F.is_pickable),
-            default=False,
-            catch=KeyErrorNotFound,
+            lambda: try_or(
```
mawildoer, line 272: Love the concept of `try_or`, but found it unwieldy, especially to read in practice.

Perhaps they'd be clearer w/ a partial? I've played with making a `try_or` with a signature something like `partial`:
```python
def try_or[T](func: Callable[[*A, **K], T], *A, excepts: type[Exception], default: T = None):
```
Might be clearer using partials rather than lambdas?

Ideally though, if nesting is going beyond a single walrus-stmt, I'd suggest we make a true function rather than a lambda for the key/filter/matcher. These deep, deep nested lambdas are also pretty hard for at least me, and I expect EEs and others getting used to the core to follow.

mawildoer (approved):

mawildoer:
Merging on your behalf, since you're OM

PR #759, opened 2024-12-24
Save failed regression artifacts

mawildoer:
The `ato create project` changes are in there because I needed `robustly_rm_dir`


pyproject.toml
```diff
@@ -102,6 +102,7 @@ dev = [
     "pytest-history @ git+https://github.com/atopile/pytest-history.git",
     "mkdocstrings[python]>=0.27.0",
     "griffe-typingdoc>=0.2.7",
+    "pathvalidate>=3.2.1",
```
mawildoer, line 105: We should use this more places!

PR #761, opened 2024-12-27
CLI: Logging upgrades

sam-mellor:
- Enables Markdown support in logging / exception messages (can be disabled with `markdown=False`)
- Marks app root node as root before visiting children, so memory address doesn't show in logs
- Make picker progress bar transient to reduce clutter
- Adds custom highlighter class with support for relative filesystem paths
- Upgrades various log/error messages

### DIFF
--- a/src/atopile/front_end.py
+++ b/src/atopile/front_end.py
@@ -1126,8 +1126,8 @@ class Bob(BasicsMixin, SequenceMixin, AtoParserVisitor):  # type: ignore  # Over
                     with downgrade(DeprecatedException), self._suppression_connect:
                         raise DeprecatedException.from_ctx(
                             ctx,
-                            f"Connected `{a.get_full_name()}` to `{b.get_full_name()}`"
-                            " by duck-typing. They should be of the same type.",
+                            f"Connected `{a}` to `{b}` by duck-typing."
+                            "They should be of the same type.",
                             traceback=self.get_traceback(),
                         )
 


src/atopile/front_end.py
```diff
@@ -833,27 +836,27 @@ def _new_node(
...
         # This should never happen
-        raise ValueError(f"Unknown item type {item}")
+        raise ValueError(f"Unknown item type `{item}`")
 
+    @contextmanager
     def _init_node(
-        self,
-        node_type: ap.BlockdefContext | Type[L.Node],
-    ) -> L.Node:
+        self, node_type: ap.BlockdefContext | Type[L.Node]
```
mawildoer, line 843: @iopapamanoglou this is an interesting pattern. Not 100% one way or another on how much I like it, but seems like it or `super()` calls could help in fabll too?

src/faebryk/libs/exceptions.py
```diff
@@ -235,7 +255,7 @@ def nom_nom_nom(self, exc: T, original_exinfo):
             exceptions = [exc]
 
         for e in exceptions:
-            self.logger.log(self.to_level, str(e), exc_info=e)
+            self.logger.log(self.to_level, str(e), exc_info=e, extra={"markdown": True})
```
mawildoer, line 258: Does this need the same tuple treatment as https://github.com/atopile/atopile/pull/761/files#diff-269e00d53db17645671d77c768728e2cb6240229264b966225af3886c81b2751R40?

src/atopile/front_end.py
```diff
@@ -1121,8 +1126,8 @@ def _connect(
                     with downgrade(DeprecatedException), self._suppression_connect:
                         raise DeprecatedException.from_ctx(
                             ctx,
-                            f"Connected {a} to {b} by duck-typing."
-                            " They should be of the same type.",
+                            f"Connected `{a.get_full_name()}` to `{b.get_full_name()}`"
```
mawildoer, line 1129: These two are going opposite directions: https://github.com/atopile/atopile/pull/761/files#diff-894ba2f8e089bd2b12321656b065fbcff36135c2e79d0d32f207314dead6de54R301

mawildoer (approved):

PR #780, opened 2025-01-02
Fix net name collisions

mawildoer:
Net names were colliding, which was causing all varieties of strange behaviour in KiCAD. 

Fix that

### DIFF
--- a/src/faebryk/exporters/netlist/graph.py
+++ b/src/faebryk/exporters/netlist/graph.py
@@ -216,6 +216,11 @@ def _name_shittiness(name: str | None) -> float:
     if name.startswith("_"):
         return 0.2
 
+    # "hv" is common from power interfaces, but
+    # if there's something else available, prefer that
+    if name == "hv":
+        return 0.3
+
     # Anything with a trailing number is
     # generally less interesting
     if re.match(r".*\d+$", name):
--- a/src/faebryk/exporters/netlist/graph.py
+++ b/src/faebryk/exporters/netlist/graph.py
@@ -287,16 +287,13 @@ def attach_net_names(nets: Iterable[F.Net]) -> None:
             net.add(F.has_overriden_name_defined(net_required_names.pop()))
             continue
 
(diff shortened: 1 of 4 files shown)


src/faebryk/exporters/netlist/graph.py
```diff
@@ -287,24 +287,22 @@ def _decay(depth: int) -> float:
...
+        names[net] = _NetName()
+
         if net_suggested_names:
-            net.add(
-                F.has_overriden_name_defined(
-                    min(net_suggested_names, key=lambda x: x[1])[0]
-                )
-            )
-            continue
+            names[net].base_name = min(net_suggested_names, key=lambda x: x[1])[0]
```
mawildoer, line 294: Instead of overriding a name directly, just take it as the winning suggestion

src/faebryk/library/ElectricPower.py
```diff
@@ -120,5 +120,4 @@ def __preinit__(self) -> None:
             F.is_bus_parameter(reduce=(self.max_current, Add))
         )
 
-        self.hv.add(F.has_net_name("vcc"))
```
mawildoer, line 123: This was troublesome, because "vcc" is a kinda mid name, but "gnd" is a must-have

src/faebryk/exporters/netlist/graph.py
```diff
@@ -287,24 +287,22 @@ def _decay(depth: int) -> float:
...
-        if implicit_name_candidates:
+        elif implicit_name_candidates:
             # Type ignored on this because they typing on both max and defaultdict.get
             # is poor. This is actually correct, and supposed to return None sometimes
             names[net].base_name = max(
                 implicit_name_candidates,
                 key=implicit_name_candidates.get,  # type: ignore
             )
 
-    # Resolve as many conflict as possible by prefixing on the lowest common node's full name # noqa: E501  # pre-existing
```
mawildoer, line 307: Fix E501

PR #788, opened 2025-01-04
Bugfix: faebryk: Library: Set weak owner

ruben-iteng:
# Bugfix: faebryk: Library: Set weak owner

## Description

Also fix reversed resistor values in one module

Fixes # (issue)

# Checklist

### DIFF
--- a/src/faebryk/library/CH344.py
+++ b/src/faebryk/library/CH344.py
@@ -70,8 +70,6 @@ class CH344(Module):
         self.gpio[14].connect(self.uart[0].dsr)
         self.gpio[15].connect(self.uart[1].dcd)
 
-        self.test.set_weak(on=False, owner=self)
-
         # ------------------------------------
         #          parametrization
         # ------------------------------------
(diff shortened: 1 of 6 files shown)


src/faebryk/library/CH344.py
```diff
@@ -70,8 +70,7 @@ def __preinit__(self):
         self.gpio[14].connect(self.uart[0].dsr)
         self.gpio[15].connect(self.uart[1].dcd)
 
-        # FIXME
-        # self.test.set_weak(on=False)
+        self.test.set_weak(on=False, owner=self)
```
mawildoer, line 73: This one is a little more complex because when `self` is picked, internal parts (in this case, including the pull resistor) won't be picked, meaning it'll be missing.

I'm not immediately sure what a good solution to this one is, unfortunately. Whatever is initing the `CH344` is probably the best candidate

mawildoer (commented):
Thanks again Ruben!

src/faebryk/library/CH344.py
ruben-iteng: Good catch, I'll move it to the reference design.
We might want to add some checks and/or design guidelines, to prevent this from happening :sweat_smile:

mawildoer: You're right. Good idea! Created this one for it: https://github.com/atopile/atopile/issues/790

napowderly (approved):

PR #792, opened 2025-01-07
fix net numbering and pad rotations

napowderly:
# Fix: Net numbering and pad rotations

# Description
Pads were not having rotation re-applied when reloaded by ato
Net re-numbering will now find next available number, avoiding duplicates.

Fixes # (issue)

# Checklist

### DIFF
--- a/src/faebryk/exporters/pcb/kicad/pcb.py
+++ b/src/faebryk/exporters/pcb/kicad/pcb.py
@@ -286,11 +286,11 @@ class PCB:
         # Add new nets -----------------------------------------------------------------
         logger.debug(f"New nets: {nets_added}")
         existing_net_numbers = {net.number for net in pcb.kicad_pcb.nets}
+        net_numbers = iter(yield_missing(existing_net_numbers))
 
         for net_name in nets_added:
             # Find the first unused net number
-            net_number = next(yield_missing(existing_net_numbers))
-            existing_net_numbers.add(net_number)
+            net_number = next(net_numbers)
 
             pcb_net = C_kicad_pcb_file.C_kicad_pcb.C_net(
                 name=net_name,
--- a/src/faebryk/exporters/pcb/kicad/pcb.py
+++ b/src/faebryk/exporters/pcb/kicad/pcb.py
@@ -26,6 +26,7 @@ from faebryk.libs.sexp.dataclass_sexp import get_parent
 from faebryk.libs.util import (
     KeyErrorNotFound,
     dataclass_as_kwargs,
+    duplicates,
     find,
     find_or,
     not_none,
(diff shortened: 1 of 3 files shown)


src/faebryk/exporters/pcb/kicad/pcb.py
```diff
@@ -283,22 +284,27 @@ def get_property_uuid(
...
         logger.debug(f"New nets: {nets_added}")
+        existing_net_numbers = {net.number for net in pcb.kicad_pcb.nets}
+
         for net_name in nets_added:
-            # nl_net = nl_nets[net_name]
-            if removed_net_numbers:
-                net_number = removed_net_numbers.pop()
-            else:
-                net_number = len(pcb.kicad_pcb.nets)
+            # Find the first unused net number
```
mawildoer, line 290: We do this a lot. While there's not currently a function for it in `util`, there probably should be

Something like (written freehand)
```python
def yield_missing(existing: Container, candidates: Iterable | None = None):
    for c in candidates:
        if c not in existing:
            yield c
```

The you can:
```suggestion
        available_numbers = iter(yield_missing(existing_net_numbers))
        for net_name in nets_added:
            net_number = next(available_numbers)
            
            # .... continue
            pcb_net = C_kicad_pcb_file.C_kicad_pcb.C_net(
                name=net_name,
                number=net_number,
            )
```

Notes:
- Candidates is `None` by default to default to a `count()`. We can't do this in the function declaration, because it won't be unique to each call
- You use it outside the for loop because then it preserves the start position to continue to the next number
- Ideally you use a set as the `existing` thing, because it's faster, but we don't need to

src/faebryk/exporters/pcb/kicad/pcb.py
```diff
@@ -432,7 +438,13 @@ def get_property_uuid(
                             if p.name in pads
                             else None,
                             # rest of fields
-                            **dataclass_as_kwargs(p),
+                            # **dataclass_as_kwargs(p),
```
mawildoer, line 441: Crud

src/faebryk/exporters/pcb/kicad/pcb.py
```diff
@@ -432,7 +438,13 @@ def get_property_uuid(
...
+                            **{
+                                k: v
+                                for k, v in dataclass_as_kwargs(p).items()
+                                if k != "at"
+                            },
+                            at=C_xyr(x=p.at.x, y=p.at.y, r=p.at.r + at.r),
```
mawildoer, line 447: This is better IMO

```suggestion
                            **{
                                **dataclass_as_kwargs(p).items()
                                # Override `at` because we need to add the part's rotation to the pads too
                                "at": C_xyr(x=p.at.x, y=p.at.y, r=p.at.r + at.r),
                            },
```

mawildoer (approved):

PR #793, opened 2025-01-08
Update CNAME record

mawildoer:


mawildoer:
Bypassed the workflows because I knocked out the site, so this is relatively time-critical

PR #794, opened 2025-01-08
Inductor current params

napowderly:
# Lib: Inductor current parameters

inductor std lib currently has only 'max_current' parameter for constraining current, which is not a standard/helpful parameter.

Inductor current is typically described by 'saturation current' and 'rated current':

Saturation Current is the current level at which an inductor’s inductance drops to a specified percentage (commonly 70% to 90%) of its nominal value due to magnetic saturation of the core material.

Rated Current is the maximum continuous current an inductor can carry without exceeding its specified temperature rise, typically due to copper and core losses.

Digikey:
Screenshot 2025-01-08 at 10 51 51 AM
JLCPCB:
Screenshot 2025-01-08 at 10 52 05 AM
Typical datasheet:
Screenshot 2025-01-08 at 10 52 40 AM

Fixes # (issue)

# Checklist

### DIFF
--- a/src/faebryk/library/Inductor.py
+++ b/src/faebryk/library/Inductor.py
@@ -23,21 +23,28 @@ class Inductor(Module):
         soft_set=L.Range(100 * P.kHz, 1 * P.GHz),
         tolerance_guess=10 * P.percent,
     )
-    current_rating = L.p_field(
+    """Frequency where inductance and parasitic capacitance cancel causing resonance."""
+
+    rated_current = L.p_field(
         units=P.A,
         likely_constrained=True,
         soft_set=L.Range(1 * P.mA, 100 * P.A),
     )
+    """Maximum continuous RMS current rating"""
+
     saturation_current = L.p_field(
         units=P.A,
         likely_constrained=True,
         soft_set=L.Range(1 * P.mA, 100 * P.A),
     )
+    """Current at which inductance drops 70% due to core saturation"""
+
     dc_resistance = L.p_field(
         units=P.Ω,
         soft_set=L.Range(10 * P.mΩ, 100 * P.Ω),
         tolerance_guess=10 * P.percent,
     )
+    """DC resistance of the inductor"""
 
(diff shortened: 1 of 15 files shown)


iopapamanoglou (changes requested):
Saturation current is fine.
But rated current is a very component-perspective way of thinking of inductors. 
In the long-term we just have current, but for now only the max of it is interesting, thus max_current.
Also consistent with capacitor max_voltage, resistor max_power etc.
If we need to track the rated part of a component separately we can add it as additional parameter with constrain.

E.g 'rated_current >=! max_current'

napowderly:
open to keeping max_current and adding rated current as a separate parameter.

The examples of max_power and max_voltage are the same issue for sure, I would also argue the same way, that as a designer you should be able to explicitly constrain at a component level, like you would a digikey search. We should add these explicitly as rated_power and rated_voltage.

For all of these examples, it makes sense to me that in the future we want to be able to add further automation, where these search criteria are derived from our circuit model, eg to determine a suitable rated_voltage = max(applied_voltage) * derating_factor.

src/faebryk/library/Inductor.py
```diff
@@ -23,16 +23,28 @@ class Inductor(Module):
...
     )
-    max_current = L.p_field(
+    """Frequency where inductance and parasitic capacitance cancel causing resonance."""
+
+    rated_current = L.p_field(
         units=P.A,
         likely_constrained=True,
         soft_set=L.Range(1 * P.mA, 100 * P.A),
     )
+    """Maximum continuous RMS current rating"""
```
mawildoer, line 33: Maayyyybe this warrants a spec attachment? Eg. the standard temp rise or whatever they spec it to

mawildoer (approved):
@iopapamanoglou , yep, understood this is a very component oriented perspective, however:
- it's the only perspective we support for now, since `max_current` isn't propagated anywhere else
- is consistent with what EEs are expecting
- carries additional attributes like "RMS" and "continuous" which are standardised

iopapamanoglou:
@mawildoer your points don't contradict my suggestion as far as I see. Or am I missing something?

PR #801, opened 2025-01-10
New config system

sam-mellor:
Switches to a new config system, using `pydantic-settings`.

- simplifies config class hierarchy
- everything is now accessed from a single global config object, including the currently-active build config
- values are sourced  by layering of CLI args, environment, project config file, and global config file

## Config sources

`ConfigFlag`
- used for development purposes only
- prefixed with `FBRK_` (for now)
- declared at point-of-use

`config` object
- represents user-facing settings hierarchy (in order of precedence)
    - specific CLI args
    - env vars (specified like `ATO_PCBNEW_AUTO=1`)
    - project-level config (`ato.yaml`)
    - global config (e.g. `~/.config/atopile/config.yaml`)
    
##  TODO
- move more CLI arg management / validation into config object
- disallow fields like `builds` in global config
- update docs

### DIFF
--- a/src/atopile/config.py
+++ b/src/atopile/config.py
@@ -12,6 +12,7 @@ from typing import Any, Callable, Generator, Iterable, Self
 from pydantic import (
     AliasChoices,
     BaseModel,
+    ConfigDict,
     Field,
     ValidationError,
     ValidationInfo,
@@ -93,6 +94,10 @@ def _try_construct_config[T](
         raise UserConfigurationError(f"Invalid config: {ex}") from ex
 
 
+class BaseConfigModel(BaseModel):
+    model_config = ConfigDict(use_attribute_docstrings=True)
+
+
 class BuildType(Enum):
     ATO = "ato"
     PYTHON = "python"
@@ -157,7 +162,7 @@ class ProjectConfigSettingsSource(ConfigFileSettingsSource):
         return self.yaml_data or {}
 
 
-class ProjectPaths(BaseModel):
+class ProjectPaths(BaseConfigModel):
     root: Path
     """Project root directory (where the ato.yaml file is located)"""
 
@@ -213,7 +218,7 @@ class ProjectPaths(BaseModel):
         self.layout.mkdir(parents=True, exist_ok=True)
 
 
-class BuildPaths(BaseModel):
+class BuildPaths(BaseConfigModel):
     layout: Path
     """Build layout file"""
 
@@ -293,7 +298,7 @@ class BuildPaths(BaseModel):
         return True
 
 
-class BuildConfig(BaseModel):
+class BuildConfig(BaseConfigModel):
     _project_paths: ProjectPaths
 
     name: str
@@ -357,7 +362,7 @@ class BuildConfig(BaseModel):
         return address.entry_section
 
 
-class Dependency(BaseModel):
+class Dependency(BaseConfigModel):
     name: str
     version_spec: str | None = None
     link_broken: bool = False
@@ -387,12 +392,12 @@ class Dependency(BaseModel):
         return str(path) if path else None
 
 
-class ServicesConfig(BaseModel):
-    class Components(BaseModel):
+class ServicesConfig(BaseConfigModel):
+    class Components(BaseConfigModel):
         url: str = Field(default="https://components.atopileapi.com")
         """Components URL"""
 
-    class Packages(BaseModel):
+    class Packages(BaseConfigModel):
         url: str = Field(default="https://get-package-atsuhzfd5a-uc.a.run.app")
         """Packages URL"""
 
@@ -407,7 +412,7 @@ class ServicesConfig(BaseModel):
     packages: Packages = Field(default_factory=Packages)
 
 
-class ProjectConfig(BaseModel):
+class ProjectConfig(BaseConfigModel):
     """Project-level config"""
 
     ato_version: str = Field(
@@ -532,10 +537,11 @@ class ProjectSettings(ProjectConfig, BaseSettings):  # FIXME
     """
 
     # TOOD: ignore but warn for extra fields
-    model_config = SettingsConfigDict(
+    model_config = BaseConfigModel.model_config | SettingsConfigDict(
         env_prefix=ENV_VAR_PREFIX,
         env_nested_delimiter="_",
         enable_decoding=False,
+        use_attribute_docstrings=True,
         extra="forbid",
     )
 


src/atopile/buildutil.py
```diff
@@ -102,13 +94,20 @@ def build(build_ctx: BuildContext, app: Module) -> None:
...
     # Pickers ------------------------------------------------------------------
-    if build_ctx.keep_picked_parts:
+    if config.build.keep_picked_parts:
         load_descriptive_properties(G)
     try:
         pick_part_recursively(app, solver)
     except PickError as ex:
         raise UserPickError.from_pick_error(ex) from ex
 
+    ensure_footprint_lib(
```
mawildoer, line 104: This probably warrants a comment while we remember why it's here

src/atopile/buildutil.py
```diff
@@ -335,9 +338,8 @@ def consolidate_footprints(build_ctx: BuildContext, app: Module) -> None:
     elif lib_prefix_on_ids and not lib_in_fptable:
         # we need to add a lib entry pointing to the footprints dir
         ensure_footprint_lib(
-            build_ctx.paths,
             "lib",
-            build_ctx.paths.build / "footprints" / "footprints.pretty",
+            config.project.paths.build / "footprints" / "footprints.pretty",
```
mawildoer, line 342: We should probably create some concept of versioned/persistent vs. ephemeral/temp/build footprints

src/atopile/cli/build.py
```diff
@@ -41,129 +43,122 @@ def build(
...
     from atopile.config import BuildType
     from faebryk.library import _F as F
     from faebryk.libs.exceptions import accumulate, log_user_errors
-    from faebryk.libs.picker import lcsc
 
-    build_ctxs = create_build_contexts(entry, build, target, option, standalone)
+    parse_build_options(entry, selected_builds, target, option, standalone)
 
-    for build_ctx in build_ctxs:
+    for _, build_cfg in config.project.builds.items():
```
mawildoer, line 54:
```suggestion
    for build_cfg in config.project.builds.values():
```

src/atopile/cli/build.py
```diff
@@ -41,129 +43,122 @@ def build(
...
-                # TODO: these should be drawn from the buildcontext like everything else
-                lcsc.BUILD_FOLDER = build_ctx.paths.build
-                lcsc.LIB_FOLDER = build_ctx.paths.component_lib
-                lcsc.LIB_FOLDER.mkdir(exist_ok=True, parents=True)
-                lcsc.KICAD_PROJECT_PATH = build_ctx.paths.kicad_project.parent
-
-                # TODO: add a mechanism to override the following with custom build machinery # noqa: E501  # pre-existing
-                buildutil.build(build_ctx, app)
+        for build in config.builds:
+            with build:
```
mawildoer, line 80: nit; dedent this block by stacking w/ the other contexts on `:82`

src/atopile/cli/cli.py
```diff
@@ -105,6 +106,20 @@ def cli(
 app.command()(view.view)
 
 
+@app.command(hidden=True)
+def export_config_schema():
```
mawildoer, line 110: nit; more accurate?

```suggestion
def dump_config_schema():
```

src/atopile/cli/build.py
```diff
@@ -77,42 +75,21 @@ def build(
...
+                    case _:
+                        raise ValueError(
+                            f"Unknown build type: {config.build.build_type}"
                         )
 
-            manifest_path = config.project.paths.manifest
-            manifest_path.parent.mkdir(exist_ok=True, parents=True)
-            with open(manifest_path, "w", encoding="utf-8") as f:
-                json.dump(manifest, f)
+                # TODO: add a mechanism to override the following with custom build machinery # noqa: E501  # pre-existing
```
mawildoer, line 91: nit; fix pre-existing noqas when touching the line

src/atopile/config.py
```diff
@@ -1,387 +1,309 @@
...
+            case "Darwin" | "Linux":
+                config_dir = (
+                    Path(os.environ.get("XDG_CONFIG_HOME", Path.home() / ".config"))
+                    / APPLICATION_NAME
+                )
+            case "Windows":
+                # TODO: @windows
+                raise UserException(f"Unsupported platform: {platform.system()}")
+            case _:
+                raise UserException(f"Unsupported platform: {platform.system()}")
```
mawildoer, line 131: I'd typically advocate for the simplest solution, and use it even on OSs that less often follow this pattern. Even `ssh` does under Windows, if I remember correctly. Which means there's precedence for forcing it

```suggestion
        config_dir = Path.home() / ".config" / APPLICATION_NAME
```

src/atopile/config.py
```diff
@@ -1,387 +1,309 @@
...
-    """
-    Get the project config from a path.
-    """
-    project_dir = get_project_dir_from_path(path)
-    project_config_file = project_dir / CONFIG_FILENAME
-    if project_config_file not in _loaded_configs:
-        _loaded_configs[project_config_file] = ProjectConfig.load(project_config_file)
-    return _loaded_configs[project_config_file]
+    def get_data(self) -> dict[str, Any]:
+        return self.yaml_data if self.yaml_data else {}
```
mawildoer, line 158: nit; simplifies to `self.yaml_data or {}` for falsey conditional

src/atopile/config.py
```diff
@@ -1,387 +1,309 @@
...
+        data.setdefault("manifest", data["build"] / "manifest.json")
+        data.setdefault("component_lib", data["build"] / "kicad" / "libs")
+        data.setdefault("modules", data["root"] / ".ato" / "modules")
+        super().__init__(**data)
+
+    @model_validator(mode="after")
+    def make_paths_absolute(model: "ProjectPaths") -> "ProjectPaths":
+        """Make all paths absolute relative to the project root."""
+        for field_name, field_value in model:
+            if field_name != "root" and isinstance(field_value, Path):
```
mawildoer, line 186: Should `root` not also be absolute?

src/atopile/config.py
```diff
@@ -1,387 +1,309 @@
...
 
-def get_project_config_from_addr(addr: address.AddrStr) -> ProjectConfig:
-    """
-    Get the project config from an address.
-    """
-    return get_project_config_from_path(Path(address.get_file(addr)))
+class ProjectPaths(BaseModel):
+    root: Path
+    src: Path
+    layout: Path
```
mawildoer, line 164: I'd be very tempted to rename this to `layout_dir`. Is there something stopping us?

src/atopile/config.py
```diff
@@ -1,387 +1,309 @@
...
-    """A class to hold the arguments to a project."""
 
-    project_path: Path  # abs path to the project directory
-    src_path: Path  # abs path to the source directory
-    module_path: Path  # abs path to the module directory
-    layout_path: Path  # eg. path/to/project/layouts/default/default.kicad_pcb
-    lock_file_path: Path  # eg. path/to/project/ato-lock.yaml
-    config: ProjectConfig
+class BuildPaths(BaseModel):
+    layout: Path
```
mawildoer, line 200: Some comments or doc-strings on these would be ideal while they're fresh. I expect we'll quickly forget about the corner-cases and discoveries you've made over the last ~4/5 days

src/atopile/config.py
```diff
@@ -1,387 +1,309 @@
...
-    )  # eg. path/to/project/<where ato.yaml is> OR git repo OR None is indiscernible
-    layout: Path  # eg. path/to/project/layouts/default/default.kicad_pcb
-    lock_file: Path | None  # eg. path/to/project/ato-lock.yaml
-    build: Path  # eg. path/to/project/build/<build-name>
-    output_base: Path  # eg. path/to/project/build/<build-name>/entry-name
-    netlist: Path
-    fp_lib_table: Path
-    component_lib: Path
-    kicad_project: Path
+class BuildEntry(BaseModel):
```
mawildoer, line 270: We typically represent these via an `AddrStr`. What's different about this case?

src/atopile/config.py
```diff
@@ -1,387 +1,309 @@
...
-    fail_on_drcs: bool
-    dont_solve_equations: bool
-    keep_picked_parts: bool
-
+    address: str | None = Field(alias="entry")
+    targets: list[str] = Field(default=["__default__"])  # TODO: validate
+    exclude_targets: list[str] = Field(default=[])
+    fail_on_drcs: bool = Field(default=False)
+    dont_solve_equations: bool = Field(default=False)
+    keep_picked_parts: bool = Field(default=False)
```
mawildoer, line 284: nano nit; now this,`keep_net_names` and `frozen` all have defaults I think they belong together

src/atopile/config.py
```diff
@@ -1,387 +1,309 @@
...
+    keep_net_names: bool = Field(default=False)
+    frozen: bool = Field(default=False)
+
+    def __init__(self, **data: Any):
+        super().__init__(**data)
+        self._project_paths = data.get("_project_paths", ProjectPaths())
+
+    @model_validator(mode="before")
+    def init_paths(cls, data: dict) -> dict:
+        match data.get("paths"):
```
mawildoer, line 295: This seems like it might be something for a temporary compatibility? If so, we should include a comment to remember such

src/atopile/config.py
```diff
@@ -391,93 +313,320 @@ def build_type(self) -> BuildType:
...
                 return BuildType.PYTHON
             case _:
-                raise atopile.errors.UserException(
-                    f"Unknown entry suffix: {suffix} for {self.entry}"
+                raise UserException(
+                    f"Unknown entry suffix: {suffix} for {self.entry_section}"
                 )
 
+    @property
+    def file_path(self) -> Path:
```
mawildoer, line 329: Perhaps warrants

```suggestion
    def entry_file_path(self) -> Path:
```

src/atopile/config.py
```diff
@@ -391,93 +313,320 @@ def build_type(self) -> BuildType:
...
+        return address.entry_section
+
+
+class Dependency(BaseModel):
+    name: str
+    version_spec: str | None = None
+    link_broken: bool = False
+    path: Path | None = None
+
+    project_config: "ProjectConfig | None" = Field(
```
mawildoer, line 345: This is interesting. Does it load the dependency's config?

test/test_config.py
```diff
@@ -1,45 +1,57 @@
...
+    dep2 = Dependency(name="esp32-s3", version_spec="^v0.0.1", path=Path("../esp32-s3"))
+
+    def add_dependency(config_data, new_data):
+        config_data["dependencies"] = config_data["dependencies"] + [new_data]
+        return config_data
+
+    config.update_project_config(add_dependency, dep1.model_dump())
+    config.update_project_config(add_dependency, dep2.model_dump())
+
+    assert config.project.dependencies is not None
```
mawildoer, line 54: Maybe out of scope of this, but does this behave gracefully with a missing `dependencies` key?

I'd guess this specific `add_dependency` function is a "probably not", but it should. It's bitten us a few times for people re-doing their deps before

src/atopile/cli/install.py
```diff
@@ -142,37 +137,38 @@ def install_single_dependency(
...
         dependency.version_spec = f"@{installed_version}"
 
-    names = {dep.name: i for i, dep in enumerate(config.dependencies)}
-    if dependency.name in names:
-        config.dependencies[names[dependency.name]] = dependency
-    else:
-        config.dependencies.append(dependency)
-    config.save_changes()
+    def add_dependency(config_data, new_data):
+        for i, dep in enumerate(config_data["dependencies"]):
```
mawildoer, line 151: May `config_data["dependencies"]` be `None` here?

src/atopile/config.py
```diff
@@ -155,18 +154,22 @@ def find_config_file(cls) -> Path | None:
...
-    root: Path
-    src: Path
-    layout: Path
-    footprints: Path
-    manifest: Path
-    build: Path
-    component_lib: Path
-    modules: Path
+    root: Path = Field(
+        description="Project root directory (where the ato.yaml file is located)"
```
mawildoer, line 162: Wonder if Pydantic plays nice w/ attr doc-strings?

src/atopile/config.py
```diff
@@ -627,6 +661,161 @@ def has_project(self) -> bool:
...
+                pass
+
+            elif not entry_arg_file_path.exists():
+                raise UserBadParameterError(
+                    "The entry you have specified does not exist.",
+                    title="Bad 'entry' parameter",
+                )
+            else:
+                raise ValueError(
+                    f"Unexpected entry path type {entry_arg_file_path} - this should never happen!"  # noqa: E501  # pre-existing
```
mawildoer, line 735: nit; E501 left

src/atopile/config.py
```diff
@@ -627,6 +661,161 @@ def has_project(self) -> bool:
...
+                    title="Bad 'entry' parameter",
+                )
+
+            entry_arg_file_path = (
+                Path(address.get_file(entry)).expanduser().resolve().absolute()
+            )
+
+        return entry, entry_arg_file_path
+
+    def apply_options(
```
mawildoer, line 762: This name makes it feel like it's in a little bit of a strange place, between something obviously CLI specific (which its function is), and something generic (perhaps something that accepts `kwargs`, applies those and cascades the update. It's fine, but wouldn't be opposed to moving it later

src/faebryk/libs/picker/api/api.py
```diff
@@ -62,25 +60,25 @@ def get_package_candidates(module: Module) -> frozenset["PackageCandidate"]:
 class ApiClient:
     @dataclass
     class Config:
-        api_url: str = API_URL.get()
-        api_key: str = API_KEY.get()
+        api_url: str = config.project.services.components.url
```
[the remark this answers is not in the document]
mawildoer: It looks like it'll be created on the first call of `_cfg`, which implies to me that it'll be a little brittle. Can eat it for now, hopefully pointing it out means we'll recall it if this ever bites us!

mawildoer (approved):

Issue #712, opened 2024-12-18
3D model path broken with ato build


iopapamanoglou:
### Current Behaviour

```
(model "${KIPRJMOD}/../libs/3dmodels/lcsc.3dshapes/R0603.wrl"
```

```
root
- standalone
  - default.kicad_pcb
- build
  - kicad
    - libs
      - 3dmodels
```

### Expected Behaviour

.

### Version

4d5ce269bc30147f102792c159515033160744f5

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

hexdae:
Hey @iopapamanoglou, maybe this is a different issue, but with `ato` v0.2.0 you could place a `.step` or `.wrl` model in the `footprints` directory and `ato build` would copy this under `build/footprints/footprints.3dshapes`

We have found that in ato 0.3.0 this is not the case anymore. Is this intended behavior or a bug?

hexdae:
I think this was the PR that introduce the behavior we relied upon https://github.com/atopile/atopile/pull/282

ruben-iteng:
Correct path: `${KIPRJMOD}/../../../build/kicad/libs/3dmodels/lcsc.3dshapes/<file_name>.wrl`
Also, idk if the footprints in `build/kicad/libs/footprints/lcsc.pretty` should contain a reference to the 3D files, they don't now

iopapamanoglou:
Reopening to check @hexdae 's issue.

Thanks @ruben-iteng for fixing this.

PR #813, opened 2025-01-10
Fix: default fp-lib-table path

ruben-iteng:
# Fix: Default fp-lib-table path

# Description

Relates to #712

Fixes 3D path
Renames BuildPaths and BuildConfig to BuildTargetxxxx to clarify.
Add Build target root

Fixes # (issue)

# Checklist


src/atopile/config.py
```diff
@@ -241,7 +241,7 @@ def __init__(self, name: str, project_paths: ProjectPaths, **data: Any):
         )
         data.setdefault("output_base", project_paths.build / name)
         data.setdefault("netlist", data["output_base"] / f"{name}.net")
-        data.setdefault("fp_lib_table", project_paths.layout.parent / "fp-lib-table")
+        data.setdefault("fp_lib_table", project_paths.layout / "fp-lib-table")
```
mawildoer, line 244:
```suggestion
        data.setdefault("fp_lib_table", project_paths.layout / "fp-lib-table")
```

mawildoer (changes requested):
Thanks Ruben! 🙌

ruben-iteng:
Also (temp) fixed 3D file path in the LCSC picker (again) :sweat_smile:

mawildoer (approved):
@sam-mellor mind taking a squizz?

hexdae:
This fixes a weird behavior we were observing as well

src/faebryk/libs/picker/lcsc.py
```diff
@@ -167,15 +164,13 @@ def download_easyeda_info(lcsc_id: str, get_model: bool = True):
...
-            else str(
+            str(
                 "${KIPRJMOD}"
                 / model_base_path_full.relative_to(
-                    config.project.paths.layout.parent, walk_up=True
+                    config.build.paths.layout, walk_up=True
                 )
             )
-            if config.project.paths.layout
+            if config.build.paths.layout
```
ruben-iteng, line 173: This line gives the follwing error when using `ato create` `> component > C34572614`

``` python
668 │   def build(self) -> BuildConfig:
669 │   │   if current := self._current_build:
670 │   │   │   return current
671 │   │   raise RuntimeError("No build config is currently active")
```
```bash
RuntimeError: No build config is currently active
```

mawildoer:
Hey @ruben-iteng ! We started on a few changes we thought this should have, paused to fix a few high-prio bugs and failed to give you a run-down on our plans before it was go-time in Europe. My apologies!

Like the other paths in `BuildConfig`, the models should just have their path in there. There should (IMO) be two sets paths: `versioned` (somewhere alongside the layout) and `ephemeral` (in the build directory)

Then, within those two groups we should have two/three paths: `footprint`, `model` (3d) and `symbol` (maybe/for later/we don't need to do it now)

The LCSC picker shouldn't be building it's own library, it should just add to the existing `ephemeral` one.

Later, `create_footprint_library` (which should really be `create_kicad_library`) can grab all the referenced footprints from those locations + the installed `modules` etc... and generate a library with all the correctly linked footprints -> models etc...

We needn't do all that in one PR; however, hopefully, it sheds a little light on the way we're thinking about this. @hexdae @LK probably interesting for you two too re. the 3d models!

Thanks again Ruben! I'll talk with Sam, but I thin we should be about good to merge this as a step on the road towards that ^

ruben-iteng:
Yes, indeed sounds like a better approach 👍 thx for the update. 
This pr should put main into a working state again so indeed merge if possible.
If you need help with stuff let me know 🚀

mawildoer (approved):

PR #818, opened 2025-01-11
ato create from lcsc component

napowderly:
# ato create from lcsc component

combining functionality of ato create --jlcpcb and libadd into one command line tool

Generates ato files that look like:

# Checklist


src/atopile/cli/create.py
```diff
@@ -548,3 +562,175 @@ def main(ctx: typer.Context):
 
 if __name__ == "__main__":
     create_app()  # pylint: disable=no-value-for-parameter
+
+
+# copied code from libadd
```
mawildoer, line 569: We should `libadd`'s copy and this comment

src/atopile/cli/create.py
```diff
@@ -548,3 +562,175 @@ def main(ctx: typer.Context):
...
+            @L.rt_field
+            def attach_via_pinmap(self):
+                return F.can_attach_to_footprint_via_pinmap(
+                    {{
+                        {gen_repeated_block(natsorted(pin_lines))}
+                    }}
+                )
+        """)
+        )
+        # template.traits.append(
```
mawildoer, line 756: Address or remove

src/atopile/cli/create.py
```diff
@@ -457,13 +471,13 @@ def component(
...
 
     # TODO: templated ato components too
-    if type_ is None:
-        type_ = ComponentType.fab
     # if type_ is None:
-    #     type_ = questionary.select(
-    #         "Select the component type", choices=list(ComponentType)
-    #     ).unsafe_ask()
-    #     assert type_ is not None
+    #     type_ = ComponentType.fab
```
mawildoer, line 475: Remove out-dated comments

src/atopile/cli/create.py
```diff
@@ -516,12 +530,14 @@ def component(
...
 
     if type_ == ComponentType.ato:
-        raise errors.UserNotImplementedError(
-            "Creating ato components are not yet supported"
-        )
+        template = AtoTemplate(name=sanitized_name, base="Module")
+        template.add_part(component)
+        out = template.dumps()
+        out_path.write_text(out)
+        rich.print(f":sparkles: Created {out_path} !")
```
mawildoer, line 537: This would be nice for both component types

src/atopile/cli/create.py
```diff
@@ -548,3 +564,244 @@ def main(ctx: typer.Context):
...
+class AtoTemplate(Template):
+    attributes: list[str] = field(default_factory=list)
+    pins: list[str] = field(default_factory=list)
+    defined_signals: set[str] = field(default_factory=set)  # Track defined signals
+
+    def add_part(self, part: Component):
+        # Basic component info
+        self.name = sanitize_name(f"{part.manufacturer_name}_{part.part_number}")
+        assert isinstance(self.name, str)
+        _, _, _, _, easyeda_symbol = download_easyeda_info(
```
mawildoer, line 601: I think this needs rebasing on `main`. This function signature has recently changed, so this will break on merge, and we don't have any tests for it

src/atopile/cli/create.py
```diff
@@ -548,3 +564,244 @@ def main(ctx: typer.Context):
...
+            [
+                f'lcsc_id = "{part.lcsc_display}"',
+                f'manufacturer = "{part.manufacturer_name}"',
+                f'mpn = "{part.part_number}"',
+                f'description = "{part.description}"',
+            ]
+        )
+
+        # # Add footprint if available
+        # if hasattr(part, 'footprint') and part.footprint:
```
mawildoer, line 616: Remove

src/atopile/cli/create.py
```diff
@@ -548,3 +564,244 @@ def main(ctx: typer.Context):
...
+            part.lcsc_display, get_model=False
+        )
+
+        # Add component metadata
+        self.attributes.extend(
+            [
+                f'lcsc_id = "{part.lcsc_display}"',
+                f'manufacturer = "{part.manufacturer_name}"',
+                f'mpn = "{part.part_number}"',
+                f'description = "{part.description}"',
```
mawildoer, line 611: We don't have a description attribute, so we should move this over to the doc-string

src/atopile/cli/create.py
```diff
@@ -548,3 +564,244 @@ def main(ctx: typer.Context):
...
+            ]
+        )
+
+        # # Add footprint if available
+        # if hasattr(part, 'footprint') and part.footprint:
+        #     self.attributes.append(f'footprint = "{part.footprint}"')
+
+        # Add datasheet if available
+        if part.datasheet_url:
+            self.attributes.append(f'datasheet = "{part.datasheet_url}"')
```
mawildoer, line 621: We don't currently have a `datasheet` attribute so we should:
1. Ideally, add one. I think it makes sense to have
2. Consider whether we should name this `datasheet_url` instead?

src/atopile/cli/create.py
```diff
@@ -548,3 +564,244 @@ def main(ctx: typer.Context):
...
+
+        # Add designator prefix from EasyEDA symbol
+        designator_prefix = easyeda_symbol.info.prefix.replace("?", "")
+        self.attributes.append(f'designator_prefix = "{designator_prefix}"')
```
mawildoer, line 625: There's a lot of this at the beginning that looks like it should be common. Something like `download_and_sanitise`

src/atopile/cli/create.py
napowderly: How do you mean?

napowderly: Determined libadd is not used elsewhere, just deleted the now duplicated functionality, will all live under ato create moving forward

src/atopile/cli/create.py
napowderly: there is a separate mechanism to generate the output between fabll and ato, fabll uses format_and_write which uses the black formatter on it, so that wont work for ato code. Tried a few ways of refactoring, but seemed more complicated than the explicit version.

mawildoer (approved):

PR #819, opened 2025-01-11
Consolidate deploy workflows

mawildoer:
## De-duplicate python wheels build and publish steps

I think we should just do this.

## Consolidate docker build and deploy w/ wheels

I'm not 100% this is a good idea. Perhaps instead we should keep another action and predicate it on the wheels build? Not sure.

See: https://stackoverflow.com/questions/58457140/dependencies-between-workflows-on-github-actions

### Update to PyPi and TestPyPi to accept wheels from `deploy.yml` workflow

Screenshot 2025-01-11 at 15 58 06

## New downside - speed

Maybe we upgrade the runner?


.github/workflows/deploy.yml
```diff
@@ -123,3 +125,62 @@ jobs:
...
+          path: dist
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ghcr.io/atopile/${{ matrix.image }}
+          flavor: |
+            latest=${{github.event_name == 'release' && github.event.action == 'published' && !github.event.release.prerelease && !github.event.release.draft}}
```
mawildoer, line 164: This is the only change from `cd.yml`

.github/workflows/deploy.yml
```diff
@@ -123,3 +125,62 @@ jobs:
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         attestations: true
+
+
+  docker:
```
mawildoer, line 130: This is almost verbatim from `cd.yml`

Issue #812, opened 2025-01-10
Hidden property not preserved for designators


napowderly:
### Current Behaviour

Pre build:
<img width="642" alt="Screenshot 2025-01-10 at 1 22 44 PM" src="https://github.com/user-attachments/assets/fcdc216e-155f-4ec2-945a-b72ba1b130dc" />

Post build:
<img width="642" alt="Screenshot 2025-01-10 at 1 22 08 PM" src="https://github.com/user-attachments/assets/2cec7e5a-2a64-4a82-b772-8d111f0aa96b" />

### Expected Behaviour

Preserves 'visible' parameter

### Possible Solution

#pr number

### Version

0.3.8.devo

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ruben-iteng:
I also experience this behaviour in all projects.

PR #824, opened 2025-01-12
Fix `apply_netlist` - 2

mawildoer:
## What?

- Move functionality to apply a design to a PCB directly into the PCB_Transformer
- Improve robustness for double-sided designs and re-linking
- Re-link using transformer methods, rather than designators

Logging for updates:

Screenshot 2025-01-11 at 19 39 07

## Remaining issues:

It appears some caps and resistors aren't being linked properly:

Screenshot 2025-01-11 at 19 38 51

## Later

- Should we keep the netlist infra as is?
- We should be more decisive about the properties we put in the layout. There's a shitload right now
- We should have a higher fidelity marker. In the schematics, I used a hash of the DFS content of the PCB (excluding the UUID/hash property itself). this was great because it meant we could see not only if we added it, but if someone modified it since then
- We should consider a `--tidy` flag that'll rip up even unmarked things from the layout. This shouldn't be on be default because it's annoying to have things like holes removed
- We should either add a netlist target, and potentially make PCB layout updates optional, which may help us ship a little faster ( @napowderly )

## Fixes:

Fixes: #812


mawildoer:
@sam-mellor , any idea why the logs with back-ticks aren't being rendered as markdown?

mawildoer:
This is built from `main` on the same repo, so there's definitely a bug introduced in this PR

Screenshot 2025-01-11 at 19 43 57

mawildoer:
> > @sam-mellor , any idea why the logs with back-ticks aren't being rendered as markdown?
> 
> `logger.info(message, extra={"markdown": True})`

Thanks Sam! Any idea why it wasn't required in the `pcb_update_logger`'s calls?

mawildoer:
Screenshot 2025-01-12 at 21 03 42
Screenshot 2025-01-12 at 21 03 39

mawildoer:
Left:

- [ ] Value property is wrong 
- [ ] `Footprint` and `footprint` are both duplicated and incorrect
- [ ] Back-side/flipped components will reset to the front side on rebuild w/ new footprint
- [ ] Designator prefix defaulting doesn't seem to work (All go to "U")
- [ ] Description is un-populated
- [ ] Net-relinking is likely brittle when splitting a large net into multiple. It won't figure out which sections of zone or trace is meant to attach to the new net

mawildoer:
> Left:
> 
> * [x]  `Value` property is wrong
> * [x]  `Footprint` and `footprint` are both duplicated and incorrect
> * [ ]  Back-side/flipped components will reset to the front side on rebuild w/ new footprint
> * [ ]  Designator prefix defaulting doesn't seem to work (All go to "U")
> * [ ]  `Description` property is empty
> * [ ]  Net-relinking is likely brittle when splitting a large net into multiple. It won't figure out which sections of zone or trace is meant to attach to the new net

mawildoer:
The last remaining issue seems to be that net names aren't kept properly!

src/faebryk/exporters/pcb/kicad/transformer.py
```diff
@@ -1310,3 +1408,515 @@ def add_git_version(
...
+            value=self._hash_lib_fp(lib_footprint),
+            uuid=self.gen_uuid(mark=True),
+        )
+
+    def insert_footprint(
+        self, lib_footprint: C_footprint, at: C_xyr | None = None
+    ) -> Footprint:
+        """Insert a footprint into the pcb, at optionally a specific position"""
+        if at is None:
+            at = copy.deepcopy(self.default_component_insert_point)
```
mawildoer, line 1447: Copy the data structure so if we later mutate it we don't end up w/ those changes everywhere

src/faebryk/exporters/pcb/kicad/transformer.py
```diff
@@ -1310,3 +1408,515 @@ def add_git_version(
...
+        lib_attrs = self._fp_common_fields_dict(lib_footprint)
+
+        lib_attrs["pads"] = [
+            C_kicad_pcb_file.C_kicad_pcb.C_pcb_footprint.C_pad(
+                net=None,
+                **{
+                    **asdict(p),
+                    # We have to handle the rotation separately because
+                    # because it must consider the rotation of the parent footprint
+                    "at": C_xyr(x=p.at.x, y=p.at.y, r=p.at.r + at.r),
```
mawildoer, line 1458: This could possibly mis-behave for flipped footprints. I pray they don't flip their coordinate system

src/faebryk/exporters/pcb/kicad/transformer.py
```diff
@@ -1310,3 +1408,515 @@ def add_git_version(
...
+            **lib_attrs,
+        )
+
+        self._set_lib_fp_hash(footprint, lib_footprint)
+
+        self.pcb.footprints.append(footprint)
+
+        return footprint
+
+    class BoardSide(StrEnum):
```
mawildoer, line 1476: Plausibly private

src/faebryk/exporters/pcb/kicad/transformer.py
```diff
@@ -1310,3 +1408,515 @@ def add_git_version(
...
+                    "at": C_xyr(x=p.at.x, y=p.at.y, r=p.at.r + footprint.at.r),
+                },
+            )
+            for p in lib_footprint.pads
+        ]
+        updates["propertys"] = {
+            **footprint.propertys,
+            **updates["propertys"],
+        }
+        for name, update in updates.items():
```
mawildoer, line 1609: This is a pretty blind "update everything that the `C_footprint` defines approach. I think this is okay, since it defines everything but there could be some weird corners w/ `unknown: CatchAll` for example

src/faebryk/exporters/pcb/kicad/transformer.py
```diff
@@ -1310,3 +1408,515 @@ def add_git_version(
...
+        hide: bool = True,
+    ) -> C_footprint.C_property:
+        return C_footprint.C_property(
+            name=property_name,
+            value=value,
+            layer=C_text_layer(layer=layer),
+            uuid=UUID(uuid),
+            effects=C_footprint.C_property.C_footprint_property_effects(
+                font=self.font,
+                hide=None,
```
mawildoer, line 1688: This probably warrants a comment. This inner `hide` should be `None` and the outer on `:1691` drives things, otherwise KiCAD will convert-on-load and you can get some weird behaviour

src/faebryk/exporters/pcb/kicad/transformer.py
```diff
@@ -1310,3 +1408,515 @@ def add_git_version(
...
+        fp_lib_path: Path,
+        logger: logging.Logger,
+        insert_point: C_xyr | None = None,
+    ) -> tuple[Footprint, bool]:
+        from faebryk.exporters.pcb.kicad.pcb import get_footprint
+
+        f_fp = component.get_trait(F.has_footprint).get_footprint()
+
+        # At this point, all footprints MUST have a KiCAD identifier
+        kicad_if_t = f_fp.get_trait(F.KicadFootprint.has_kicad_identifier)
```
mawildoer, line 1706: This is all a little iffy. The footprint types and traits could use a bit of a scrub

src/faebryk/exporters/pcb/kicad/transformer.py
```diff
@@ -1310,3 +1408,515 @@ def add_git_version(
...
+        address = component.get_full_name()
+
+        # All modules MUST have a designator by this point
+        ref = component.get_trait(F.has_designator).get_designator()
+
+        ## Update existing footprint
+        if pcb_fp_t := f_fp.try_get_trait(self.has_linked_kicad_footprint):
+            pcb_fp = pcb_fp_t.get_fp()
+            if fp_id != pcb_fp.name:
+                lib_fp = copy.deepcopy(get_footprint(fp_id, fp_lib_path))
```
mawildoer, line 1719: We need to copy these because we mutate them (`:1733`). I used deep copy to avoid issues down the line if we further mutate nested structures and it's cheap

src/faebryk/exporters/pcb/kicad/transformer.py
```diff
@@ -1310,3 +1408,515 @@ def add_git_version(
...
+                if existing_hash_prop := pcb_fp.propertys.get(self._FP_LIB_HASH):
+                    existing_hash = existing_hash_prop.value
+                else:
+                    existing_hash = None
+
+                # If the hash never existed, or it's changed then update the footprint
+                if existing_hash is None or existing_hash != self._hash_lib_fp(lib_fp):
+                    logger.info(
+                        f"Updating `{pcb_fp.name}`->`{fp_id}` on `{address}` ({ref})",
```
mawildoer, line 1728: I'm not in love with this logging format. I was considering a table, but that'd require a bit more construction. Maybe worth it?

src/faebryk/libs/app/designators.py
```diff
@@ -106,13 +106,3 @@ def _get_pcb_designator(fp_trait: PCB_Transformer.has_linked_kicad_footprint):
             node.add(F.has_designator(designator))
 
     return known_designators
-
-
-def override_names_with_designators(graph: Graph):
```
mawildoer, line 111: We now directly use the designators, rather than going via the name

src/faebryk/libs/util.py
```diff
@@ -1057,10 +1057,6 @@ def join_if_non_empty(sep: str, *args):
     return sep.join(s for arg in args if (s := str(arg)))
 
 
-def dataclass_as_kwargs(obj: Any) -> dict[str, Any]:
```
mawildoer, line 1060: This is just `dataclasses.asdict`

src/faebryk/libs/util.py
```diff
@@ -1106,19 +1102,25 @@ def __init__(self, data: Iterable[T] = tuple(), hasher: Callable[[T], H] = id):
         for item in data:
             self._deref[self._hasher(item)].append(item)
 
-    def add(self, item: T):
-        if item not in self._deref[self._hasher(item)]:
-            self._deref[self._hasher(item)].append(item)
+    def add(self, value: T):
```
mawildoer, line 1105: Changes are to `FuncSet`. I thought these type issues may have been why the `difference_update` wasn't working from `collections.abc.MutableSet`, but it just doesn't exist

iopapamanoglou:
Is this rtr or should this be back in draft? @mawildoer

mawildoer:
> Is this rtr or should this be back in draft? @mawildoer

Ready for review. I'm just exercising it further against a bunch of our projects

src/faebryk/exporters/pcb/kicad/transformer.py
```diff
@@ -1310,3 +1408,524 @@ def add_git_version(
...
+        cluster_points = defaultdict(_new_cluster)
+        cluster_points[None]  # Trigger top-level components, so they're at the top
+
+        def _increment_cluster_point(key):
+            cluster_points[key].y += VERTICAL_SPACING
+
+        for component, _ in gf.nodes_with_trait(F.has_footprint):
+            # FIXME: it'd be nice to allow query composition so
+            # we could ask for all the modules with footprints
+            assert isinstance(component, Module)
```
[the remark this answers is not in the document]
mawildoer: As I understand it, the goal of the `GraphFunctions` is to compose queries the cpp core can efficiently execute

src/atopile/buildutil.py
```diff
@@ -176,8 +173,9 @@ def _try_relative(path: Path) -> Path:
             f"Original layout: {original_relative}\n"
             f"Updated layout: {updated_relative}\n"
             "You can see the changes by running:\n"
-            f'diff --color "{original_relative}" "{updated_relative}"',
+            f'`diff --color "{original_relative}" "{updated_relative}"`',
             title="Frozen failed",
+            # No markdown=False here because we have both a command and paths
```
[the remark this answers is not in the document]
mawildoer: Far from ideal!

src/faebryk/exporters/pcb/kicad/transformer.py
```diff
@@ -330,9 +330,9 @@ def bind_footprint(self, pcb_fp: Footprint, module: Module):
                 logger.warning(f"No PCB pads for pad in design: {fpad}")
             fpad.add(self.has_linked_kicad_pad(pcb_fp, pads, self))
 
-        # Check for unlinked PCB pads
-        if pcb_pads:
-            logger.warning(f"No pads in design for PCB pads: {pcb_pads}")
```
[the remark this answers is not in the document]
mawildoer: Good idea. Done

Issue #822, opened 2025-01-12
`test_config::test_update_project_config` and `::test_roundtrip` fail locally


mawildoer:
It appears that they're detecting a project a level up and adjacent to my repo?

It'd make sense if it were purely a level up, but I'm confused as to how it makes it into `rp2040`

```
/Users/mattwildoer/Projects/atopile-workspace/atopile/test/test_config.py::test_roundtrip failed: tmp_path = PosixPath('/private/var/folders/8v/zd4990r94sj86c3cz5gszwcc0000gn/T/pytest-of-mattwildoer/pytest-17/test_roundtrip0')

    def test_roundtrip(tmp_path: Path):
        config_path = tmp_path / PROJECT_CONFIG_FILENAME
        config_path.write_text(TEST_CONFIG_TEXT)
>       config.project_dir = tmp_path

test/test_config.py:28:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/atopile/config.py:628: in project_dir
    self._project = _try_construct_config(ProjectSettings)
src/atopile/config.py:81: in _try_construct_config
    return model(**kwargs)
.venv/lib/python3.13/site-packages/pydantic_settings/main.py:171: in __init__
    super().__init__(
src/atopile/config.py:474: in add_dependencies
    Dependency.from_str(dep) if isinstance(dep, str) else Dependency(**dep)
src/atopile/config.py:375: in <lambda>
    default_factory=lambda data: ProjectConfig.from_path(data["path"]), exclude=True
src/atopile/config.py:451: in from_path
    file_contents["paths"].setdefault("root", path)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = {'ato-version': 'v0.2.9', 'builds': {'default': {'entry': 'RP2040Kit.ato:RP2040Kit'}}, 'dependencies': ['generics']}
key = 'paths'

    def __getitem__(self, key: Any) -> Any:
        try:
>           return ordereddict.__getitem__(self, key)
E           KeyError: 'paths'

.venv/lib/python3.13/site-packages/ruamel/yaml/comments.py:853: KeyError
```

PR #829, opened 2025-01-13
Examples must pass ✅

mawildoer:
Examples are un-marked as regression tests and must pass as part of regular testing.
They're also split into another file to make it clear they're no longer considered regression tests.

Additionally:
- Consolidate around the `util.repo_root` function
- Create a common `repo_root` fixture in `conftest.py`
- Rewrite `test_names` with pytest semantics


mawildoer:
Considering marking the "mcu" as slow so it doesn't run in CI, but it's not sloe enough to really bother

mawildoer:
Per last commit message @sam-mellor , ato paths are mangled. Appears perhaps related to https://github.com/atopile/atopile/issues/822

Skipping ato tests for now

mawildoer:
Made an issue for that update here too: https://github.com/atopile/atopile/issues/830

Issue #816, opened 2025-01-10
Voltage divider not solving correctly


napowderly:
### Current Behaviour

from "generics/resistors.ato" import Resistor

module VdivTest:
    voltage_input = 15V +/- 10%
    voltage_output = 5V +/- 20%	

    resistor_top = new Resistor
    resistor_bottom = new Resistor

    assert voltage_input * resistor_bottom.resistance / (resistor_top.resistance + resistor_bottom.resistance) within voltage_output

Picked parts:
Designator,Footprint,Quantity,Value,Manufacturer,Partnumber,LCSC Part #
"U1, U2",R0603,2,10kΩ ±1% 100mW,UNI-ROYAL(Uniroyal Elec),0603WAF1002T5E,C25804

The assertion statement has no effect on the resistor values, tested a handful of values for input/output voltage and always get 10kohm for resistor values.

### Expected Behaviour

Rtop: ~20kohm
Rbot: ~ 10kohm

### Possible Solution

#pr number 

### Version

0.3.8.dev0

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #836, opened 2025-01-15
Fix (some) issues solving for voltage dividers

sam-mellor:
Ref: https://github.com/atopile/atopile/issues/816

- refactoring of solver loop and graph mutations
- new simplifications: `isolate_lone_params`, `uncorrelated_alias_fold`
- deprecates `<` and `>` in favor of `<=` and `>=`
- various bug fixes

### DIFF
--- a/src/faebryk/library/Winbond_W25Q16JVUXIQ.py
+++ b/src/faebryk/library/Winbond_W25Q16JVUXIQ.py
@@ -62,6 +62,6 @@ class Winbond_W25Q16JVUXIQ(F.SPIFlash):
         # ------------------------------------
         #          parametrization
         # ------------------------------------
-        (self.memory_size < 16 * P.Mbit).constrain()
+        (self.memory_size <= 16 * P.Mbit).constrain()
 
         self.power.voltage.constrain_subset(L.Range(2.7 * P.V, 3.6 * P.V))


iopapamanoglou:
The multiply unpack fix is not right.
It's correct that the unpack made no sense. But you need to revert it to how it was before (mutator.mutate(e, new_operatands[0]). Because now it doesn't unfold it directly. Theoretically not a problem since it will happen in the next iteration, but practically then unnecessary to special case it.

iopapamanoglou:
Configurable timeout todo is still a problem. Because changing the class attribute doesn't do anything. Let's just remove the timeout class attribute completely from solvers for now to avoid confusion

iopapamanoglou:
By allowing subsets too in the subset & fold algorithms we need some extra changes because invariants break.

E.g:
- can't filter out all constrainedexprs because ss allows non-singles compared to alias

iopapamanoglou:
Feels wrong to skip all canonical predicate from the alias folds.  

A > B, A is [0, 10], B is [20, 30], A further uncorrelated B
-> [0,10] > [20, 30]

This is very important else we can't find contradictions at all.
Right now this logic happens in fold_ge, but it needs to go away there since it doesn't respect correlations.
(see notion comment in "Doing").

Constrained 'Is' is special because it tracks our aliases in the first place, thus we can't really mutate it (only duplicate.

IsSubset should imo not be filtered because if we have A is Lit, it's totally fine to replace A with Lit in an expr A ss Lit2. Independent whether it's constrained or not.

So in that sense the todo raised a valid concern (don't filter issubset, only filter constrained alias_is).

src/atopile/front_end.py
```diff
@@ -1320,10 +1318,23 @@ def visitComparison(
         predicates = []
         for (lh, rh), op_str in zip(itertools.pairwise(exprs), op_strs):
             match op_str:
+                # @v0.4 upgrade to error
                 case "<":
-                    op = LessThan
+                    with downgrade(
+                        errors.UserNotImplementedError, to_level=logging.WARNING
+                    ):
+                        raise errors.UserNotImplementedError(
```
mawildoer, line 1326: FYI, Ioannis, Sam and I had a discussion re. this and it makes the most sense on net. We're liable to add the, they're very standard and likely expected come integers, so `NotImplemented` is more accurate

iopapamanoglou (approved):

PR #859, opened 2025-01-24
Update `ato install`

mawildoer:
Link generics by default on install.
Change option for vendoring to `--vendor`
Install generics by default on `ato create`

### DIFF
--- a/src/atopile/cli/create.py
+++ b/src/atopile/cli/create.py
@@ -232,7 +232,7 @@ def project(
 
     # Install dependencies listed in the ato.yaml, typically just generics
     do_install(
-        to_install=None,
+        to_install="generics",
         link=True,
         upgrade=True,
         path=Path(repo_obj.working_tree_dir),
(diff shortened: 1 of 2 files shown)


src/atopile/cli/install.py
```diff
@@ -153,7 +156,11 @@ def add_dependency(config_data, new_data):
             config_data["dependencies"] = []
 
         for i, dep in enumerate(config_data["dependencies"]):
-            if dep["name"] == new_data["name"]:
+            if not isinstance(dep, dict):
```
mawildoer, line 159: Supports old configs w/ just raw name as entry

src/atopile/cli/create.py
```diff
@@ -232,7 +232,7 @@ def project(
 
     # Install dependencies listed in the ato.yaml, typically just generics
     do_install(
-        to_install=None,
+        to_install="generics",
```
mawildoer, line 235: Project template currently had "generics" in it.

This is a trade-off, but is required since otherwise the generics is unversioned in the project you're currently in

PR #860, opened 2025-01-24
Fix SyntaxError tracebacks and refactor error handling and token-based source tracking

mawildoer:
- Update `errors.py` to use token-based source tracking instead of context-based
- Modify `parse_utils.py` to support token-based lexer reconstruction
- Simplify error listener in `parse.py`
- Add test for syntax error handling
- Improve error message generation and source code highlighting

### DIFF
--- a/src/atopile/errors.py
+++ b/src/atopile/errors.py
@@ -111,7 +111,7 @@ class _BaseUserException(_BaseBaseUserException):
         cls: type[T],
         token_stream: CommonTokenStream,
         origin_start: Token,
-        origin_stop: Token,
+        origin_stop: Token | None,
         message: str,
         *args,
         traceback: Sequence[ParserRuleContext | None] | None = None,
--- a/src/atopile/errors.py
+++ b/src/atopile/errors.py
@@ -112,7 +112,7 @@ class _BaseUserException(_BaseBaseUserException):
         return instance
 
     @classmethod
-    def from_tokens[T](
+    def from_tokens[T: _BaseUserException](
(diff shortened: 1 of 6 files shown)


src/atopile/errors.py
```diff
@@ -1,23 +1,30 @@
...
 from faebryk.libs.exceptions import UserException as _BaseBaseUserException
 
 if TYPE_CHECKING:
     from faebryk.libs.picker.picker import PickError
 
 
-def _render_ctx(ctx: ParserRuleContext) -> list[ConsoleRenderable]:
-    lexer = PygmentsLexerReconstructor.from_ctx(ctx, 1, 1)
-    (src_path, src_line, src_col, _, _) = get_src_info_from_ctx(ctx)
+def _render_tokens(
```
mawildoer, line 21: Switching to tokens means we're running based on a lower-level object that's available for all errors

src/atopile/parse.py
```diff
@@ -14,44 +13,21 @@
 log = logging.getLogger(__name__)
 log.setLevel(logging.INFO)
 
-IMMEDIATE_RAISE = False
```
mawildoer, line 17: The collection of syntax errors seems to obfuscate the legitimate ones with both downstream errors and seems to offset their origin in code. Thus I believe it's better to take this approach and raise greedily, one at a time in this case

mawildoer:
```ato
module App:
    pass
     pass
```

Produces:

Screenshot 2025-01-24 at 16 19 32

PR #869, opened 2025-01-27
Fix `assert` ... `is` ... operator

mawildoer:


src/atopile/parse.py
```diff
@@ -47,6 +47,9 @@ def make_parser(src_stream: InputStream) -> AtoParser:
     stream = CommonTokenStream(lexer)
     parser = AtoParser(stream)
 
+    parser.removeErrorListeners()
```
mawildoer, line 50: Moved these up here so tests would get properly converted syntax errors too

src/atopile/parser/AtoParser.g4
```diff
@@ -82,7 +82,9 @@ compare_op_pair
     | gt_arithmetic_or
     | lt_eq_arithmetic_or
     | gt_eq_arithmetic_or
-    | in_arithmetic_or;
+    | in_arithmetic_or
+    | is_arithmetic_or
```
mawildoer, line 86: This missing was the core issue

test/front_end/conftest.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+from atopile.front_end import Bob
+
+
+@pytest.fixture
```
mawildoer, line 6: Moved this up here to commonise

test/front_end/test_front_end_assert.py
```diff
@@ -0,0 +1,21 @@
...
+from faebryk.libs.library import L
+
+
+def _parse(src: str) -> AtoParser:
+    input = InputStream(src)
+    input.name = "test"
+    return make_parser(input)
+
+
+def test_assert_is(bob: Bob):
```
mawildoer, line 15: I didn't have great tools at hand to actually assess the graph produced, so I left it here. In the future we should ideally extend these

napowderly (approved):

PR #870, opened 2025-01-27
Support relinking layouts without address as a property (from v0.2)

mawildoer:


pyproject.toml
```diff
@@ -180,7 +180,6 @@ exclude = [
     "src/atopile/parser/AtoParserBase.py",
     "src/atopile/parser/AtoParserListener.py",
     "src/atopile/parser/AtoParserVisitor.py",
-    "src/atopile/kicad_plugin",
```
mawildoer, line 183: There's a few more changes in here from linting the `kicad_plugin` dir too

src/atopile/kicad_plugin/common.py
```diff
@@ -48,17 +48,36 @@ def groups_by_name(board: pcbnew.BOARD) -> dict[str, pcbnew.PCB_GROUP]:
...
+
+def get_footprint_addr(fp: pcbnew.FOOTPRINT, uuid_map: dict[str, str]) -> Optional[str]:
     """Return the property "atopile_address" of a footprint."""
     field: pcbnew.PCB_FIELD = fp.GetFieldByName("atopile_address")
     if field:
         return field.GetText()
+
+    # This is a backup to support old layouts which
+    # haven't yet got addresses in their properties
+    uuid = get_footprint_uuid(fp)
```
mawildoer, line 65: This is the important part of the change: 

If we don't have an address property, then try to key it off known UUIDs instead

napowderly (approved):

PR #874, opened 2025-01-28
Picker: Ensure module parameters are compatible with component attributes

sam-mellor:
Fixes accepting incompatible components when not all expected attributes are returned.
 - Previously: attributes of component returned by API are checked against module parameters
 - Now: module parameters are checked against attributes of component returned by API
 
Makes mapping between parameters and component attribute names explicit, and colocated with the applicable module.

### DIFF
--- a/src/faebryk/library/LDO.py
+++ b/src/faebryk/library/LDO.py
@@ -72,6 +72,7 @@ class LDO(Module):
                 "output_voltage": self.output_voltage,
                 "quiescent_current": self.quiescent_current,
                 "dropout_voltage": self.dropout_voltage,
+                # TODO: add support in backend
                 # "ripple_rejection_ratio": self.ripple_rejection_ratio,
                 "output_polarity": self.output_polarity,
                 "output_type": self.output_type,
--- a/src/faebryk/library/LED.py
+++ b/src/faebryk/library/LED.py
@@ -49,7 +49,7 @@ class LED(F.Diode):
     def pickable(self):
         return F.is_pickable_by_type(
             F.is_pickable_by_type.Type.LED,
-            F.Diode().get_trait(F.is_pickable_by_type).get_parameters()  # TODO: better
+            F.Diode().get_trait(F.is_pickable_by_type).get_parameters()
             | {
                 "max_brightness": self.max_brightness,
                 "color": self.color,
--- a/src/faebryk/library/MOSFET.py
+++ b/src/faebryk/library/MOSFET.py
@@ -39,6 +39,7 @@ class MOSFET(Module):
             F.is_pickable_by_type.Type.MOSFET,
             {
                 "channel_type": self.channel_type,
+                # TODO: add support in backend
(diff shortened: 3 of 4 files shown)


src/faebryk/library/Capacitor.py
```diff
@@ -46,7 +46,16 @@ class TemperatureCoefficient(Enum):
         F.has_designator_prefix.Prefix.C
     )
 
-    pickable = L.f_field(F.is_pickable_by_type)(F.is_pickable_by_type.Type.Capacitor)
+    @L.rt_field
+    def pickable(self) -> F.is_pickable_by_type:
```
mawildoer, line 50: As an aside, I'd prefer for these to be private/dunder because they should be accessed via `get_trait`, but we can't because there's a private filter at the moment. We should consider changing that

src/faebryk/library/LDO.py
```diff
@@ -63,7 +63,21 @@ class OutputPolarity(Enum):
...
+    @L.rt_field
+    def pickable(self) -> F.is_pickable_by_type:
+        return F.is_pickable_by_type(
+            F.is_pickable_by_type.Type.LDO,
+            {
+                "max_input_voltage": self.max_input_voltage,
+                "output_voltage": self.output_voltage,
+                "quiescent_current": self.quiescent_current,
+                "dropout_voltage": self.dropout_voltage,
+                # "ripple_rejection_ratio": self.ripple_rejection_ratio,
```
mawildoer, line 75: Why's this one commented out? Probably deserves a comment in code

src/faebryk/library/LED.py
```diff
@@ -45,7 +45,16 @@ class Color(Enum):
     max_brightness = L.p_field(units=P.candela)
     color = L.p_field(domain=L.Domains.ENUM(Color))
 
-    pickable = L.f_field(F.is_pickable_by_type)(F.is_pickable_by_type.Type.LED)
+    @L.rt_field
+    def pickable(self):
+        return F.is_pickable_by_type(
+            F.is_pickable_by_type.Type.LED,
+            F.Diode().get_trait(F.is_pickable_by_type).get_parameters()  # TODO: better
```
mawildoer, line 52: I actually don't think this is too bad. Makes sense to me

src/faebryk/library/MOSFET.py
```diff
@@ -33,7 +33,19 @@ class SaturationType(Enum):
...
     )
 
-    pickable = L.f_field(F.is_pickable_by_type)(F.is_pickable_by_type.Type.MOSFET)
+    @L.rt_field
+    def pickable(self) -> F.is_pickable_by_type:
+        return F.is_pickable_by_type(
+            F.is_pickable_by_type.Type.MOSFET,
+            {
+                "channel_type": self.channel_type,
+                # "saturation_type": self.saturation_type,
```
mawildoer, line 42: Comment on why it's missing

PR #876, opened 2025-01-28
Revert attempt to harden `gh-pages`

mawildoer:


mawildoer:
Bypass to revert docs actions testing

PR #880, opened 2025-01-28
Fix syncing with nested layouts

mawildoer:
Currently, this just supports syncing the layout of the top-level module.

I implemented nesting modules and synced them as well. However, it crashed KiCAD and debugging that was impractical and not the best use of time. I've left the stubs of that work for the future

Additionally, this fixes logging and adds logs to pull, push and reload groups


napowderly (approved):
Tested on spacebms hil board

PR #895, opened 2025-01-29
Fix `(island)` zones sexp decoding, `CatchAll` and `CatchAll` reproduction sequence

### LINKED ISSUE #776: KiCad pcb file format filled zones island bugged
### Current Behaviour

KiCad throw a file error when there is a zone without net but with the `island` flag. 

### Expected Behaviour

KiCad loads the file

### Possible Solution

Flag might be at the wrong position (after pts, instead of above)

### Version

0.3.3-dev

### Relevant log output

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mawildoer:
# Description

I've managed to replicate this one and think it's actually a subtly different cause; it's a fault in the `CatchAll` and sexp de/encoder that's moving a positional `(island)` from slot 1 to the end.

I think there are three issues:

1. The positional `(island)` isn't being parsed properly
2. The decoder isn't properly rejecting additional statements without 
3. The `CatchAll` isn't re-placing values back in their correct indices

Fixes #776 

# Checklist

- [x] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works (if applicable)

### DIFF
--- a/src/faebryk/libs/kicad/fileformats.py
+++ b/src/faebryk/libs/kicad/fileformats.py
@@ -933,9 +933,8 @@ class C_kicad_pcb_file(SEXP_File):
             @dataclass(kw_only=True)
             class C_filled_polygon:
                 layer: str
-                island: Optional[bool] = field(
-                    **sexp_field(positional=True), default=None
-                )
+                # FIXME: decode (island), a bracketed positional
+                # We're currently relying on the CatchAll to re-serialise it
                 pts: C_pts
                 unknown: CatchAll = None
 
--- a/src/faebryk/libs/sexp/dataclass_sexp.py
+++ b/src/faebryk/libs/sexp/dataclass_sexp.py
@@ -370,7 +370,7 @@ def _decode[T](
                 f" {sp.assert_value} but is {value_dict[f.name]}"
             )
 
-    # Unprocessed values should be None is there aren't any,
+    # Unprocessed values should be None if there aren't any,
     # so they don't get reproduced etc...
     if catch_all_fields and not unprocessed:
         value_dict[catch_all_fields[0].name] = None
--- a/src/faebryk/libs/sexp/dataclass_sexp.py
+++ b/src/faebryk/libs/sexp/dataclass_sexp.py
@@ -1,10 +1,12 @@
+from __future__ import annotations
+
 import logging
 from dataclasses import Field, dataclass, fields, is_dataclass
 from enum import Enum, IntEnum, StrEnum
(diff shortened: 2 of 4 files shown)


src/faebryk/libs/sexp/dataclass_sexp.py
```diff
@@ -380,19 +385,21 @@ def _decode[T](
...
-            else:
-                vs = [v]
-            for v_ in vs:
-                if is_dataclass(v_):
-                    setattr(v_, "_parent", out)
-        return out
     except TypeError as e:
         raise TypeError(f"Failed to create {t} with {value_dict}") from e
 
+    # set parent pointers for all dataclasses in the tree
```
mawildoer, line 391: Moved outside the `try` for a more rigorous exception as to why it's failing

src/faebryk/libs/sexp/dataclass_sexp.py
```diff
@@ -171,7 +176,7 @@ def _convert(
 netlist_type = list[netlist_obj]
 
 
-def _decode[T](
+def _decode[T: DataclassInstance](
```
mawildoer, line 179: Fixes some type errors

src/faebryk/libs/kicad/fileformats.py
```diff
@@ -933,9 +933,8 @@ class E_keepout_bool(SymEnum):
             @dataclass(kw_only=True)
             class C_filled_polygon:
                 layer: str
-                island: Optional[bool] = field(
-                    **sexp_field(positional=True), default=None
-                )
+                # FIXME: decode (island), a bracketed positional
```
mawildoer, line 936: I don't like this, however as far as I can tell, don't have a good mechanism to decode the presence/absence of a positional bracketed field (eg. `(island)`)

We also don't use it, so we don't need a great output from it

src/faebryk/libs/sexp/dataclass_sexp.py
```diff
@@ -486,8 +493,8 @@ def _append_kv(name, v):
 
     if catch_all_field is not None:
         catch_all: dict[int, Any] = getattr(t, catch_all_field.name, {}) or {}
-        for i, v in catch_all.items():
-            sexp.insert(i + 1, v)
+        for i, v in sorted(catch_all.items(), key=lambda x: x[0]):
```
mawildoer, line 496: This is pre-sorted, but I like the guarentee

PR #896, opened 2025-01-29
fix error message for ERC short

napowderly:
Power ERC short check currently passes a power interface to the ercfault short exception handler, should pass hv and lv.


mawildoer (approved):

mawildoer:
Stil imperfect, since it doesn't really tell you where to look. The `Path` returned by the search algorithm seems like it'd be extremely useful here

PR #898, opened 2025-01-29
Add shim for NFET and PFET

napowderly:
# MOSFET Shims

Added shim for N and P ch mosfets

### DIFF
--- a/src/atopile/attributes.py
+++ b/src/atopile/attributes.py
@@ -280,6 +280,22 @@ def _handle_package_shim(module: L.Module, value: str, starts_with: str):
         module.add(F.has_package(pkg))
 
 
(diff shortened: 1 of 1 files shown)


src/atopile/attributes.py
```diff
@@ -280,6 +280,22 @@ def _handle_package_shim(module: L.Module, value: str, starts_with: str):
         module.add(F.has_package(pkg))
 
 
+@_register_shim("generics/mosfets.ato:NFET", "import NFET")
+class NFET(F.MOSFET):
+    """Temporary shim to translate NFETs."""
+
+    def __postinit__(self):
```
mawildoer, line 287: Should the standard lib also have separate N and P FET types?

napowderly: Quite possibly

PR #899, opened 2025-01-29
Fix footprint lib creation with a build with dir spec'd

mawildoer:
@sam-mellor his the same issue you were, and managed to fix it

Screenshot 2025-01-29 at 14 46 47

### DIFF
--- a/src/faebryk/libs/app/pcb.py
+++ b/src/faebryk/libs/app/pcb.py
@@ -231,7 +231,7 @@ def create_footprint_library(app: Module) -> None:
 
                 else:
                     try:
-                        path = path.relative_to(config.project.paths.root)
+                        prj_rel_path = path.relative_to(config.project.paths.root)
 
                     # Raised when the file isn't relative to the project directory
                     except ValueError as ex:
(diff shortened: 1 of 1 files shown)


src/faebryk/libs/app/pcb.py
```diff
@@ -244,16 +244,17 @@ def create_footprint_library(app: Module) -> None:
...
-                        mini_hash = hash_string(str(path))[:6]
-                        id_ = f"{LIB_NAME}:{path.stem}-{mini_hash}"
+                        mini_hash = hash_string(str(prj_rel_path))[:6]
+                        id_ = f"{LIB_NAME}:{prj_rel_path.stem}-{mini_hash}"
                         new_path = (
-                            atopile_fp_dir / f"{path.stem}-{mini_hash}{path.suffix}"
+                            atopile_fp_dir
+                            / f"{prj_rel_path.stem}-{mini_hash}{prj_rel_path.suffix}"
                         )
                         shutil.copy(path, new_path)
```
mawildoer, line 256: This needs to be absolute

PR #901, opened 2025-01-30
Use parameter sets, rather than util functions to configure the F.PoweredLED

mawildoer:


iopapamanoglou (changes requested):
This just seems to assume a bunch of things being unused and stashes them together.
Which is not the case. Connecting poweredleds via high side is something that a bunch of projects use for example. We can instead of removing things, just add a more simple poweredled if that is something that is desired.

mawildoer:
> This just seems to assume a bunch of things being unused and stashes them together. Which is not the case. Connecting poweredleds via high side is something that a bunch of projects use for example. We can instead of removing things, just add a more simple poweredled if that is something that is desired.

mawildoer:
> This just seems to assume a bunch of things being unused and stashes them together. Which is not the case. Connecting poweredleds via high side is something that a bunch of projects use for example. We can instead of removing things, just add a more simple poweredled if that is something that is desired.

The functionality still exists via parameter's constraints - which is how we expect things to be done. We need to pare down the unused functions in the standard library and upgrade them to the recommended best practices.

If there's an issue for a specific user with this, we can consider deprecating them first.

ruben-iteng:
The lowside resistor functionality in `PoweredLED` is handy for PCB layout flexibility. Is there another way to get this functionality?
I'm pro making the PoweredLED module better as it is/will be one of the most used modules 😄

mawildoer:
I believe the concerns are all addressed @iopapamanoglou

PR #906, opened 2025-01-30
[RFC] local deps

hexdae:
# feat: add local dependencies

# Description

This PR adds the ability to install dependences from local folders. This is particularly useful for monorepo setups.
```bash
├── components
│   ├── AT24C02D-XHM-B
│   │   ├── AT24C02DXHMB.ato
│   │   ├── ato.yaml <---------|
├── modules                    |
│   ├── at24c02d               |
│   │   ├── at24c02d.ato       |
│   │   ├── ato.yaml ----------|
```

This is an example of ato.yaml, generated by the following command

```bash
ato install --local ../../components/AT24C02D-XHM-B
```

```yaml
ato-version: ^0.3.0
builds:
  default:
    entry: at24c02d.ato:At24c02d
dependencies:
- name: AT24C02D-XHM-B
  version_spec:
  link_broken: false
  path: .ato/modules/AT24C02D-XHM-B
  source:
     local: ../../components/AT24C02D-XHM-B
```

# Checklist

- [x] I have made corresponding changes to the documentation (if applicable)
- [x] I have added tests that prove my fix is effective or that my feature works (if applicable)


src/atopile/config.py
```diff
@@ -409,9 +410,10 @@ class Dependency(BaseConfigModel):
     version_spec: str | None = None
     link_broken: bool = False
     path: Path | None = None
+    local: Path | None = None
 
     project_config: "ProjectConfig | None" = Field(
-        default_factory=lambda data: ProjectConfig.from_path(data["path"]), exclude=True
+        default_factory=lambda data: ProjectConfig.from_path(data["local"] if data["local"] else data["path"]), exclude=True
```
mawildoer, line 416: I believe this should remain pointing at the `data["path"]` in all cases. There should be a local copy of the dependency available to load this information from. If there isn't, it should be copied, rather than optionally pointing elsewhere

src/atopile/cli/install.py
```diff
@@ -198,6 +224,14 @@ def install_project_dependencies(upgrade: bool):
                     raise
 
 
+def install_local_dependency(dependency: Dependency):
+    src = dependency.local
+    dst = dependency.path or config.project.paths.modules / dependency.name
+    if not src.exists():
+        raise errors.UserException(f"Local dependency path {src} does not exist")
```
mawildoer, line 231: We can narrow this to a `UserFileNotFoundError`

src/atopile/cli/install.py
```diff
@@ -39,6 +40,9 @@ def install(
             help="Copy the contents of this dependency into the repo",
         ),
     ] = False,
+    local: Annotated[
```
mawildoer, line 43: While I'm suggesting we move the config structure a layer lower, I think this is okay to stay as is

src/atopile/config.py
```diff
@@ -409,9 +410,10 @@ class Dependency(BaseConfigModel):
     version_spec: str | None = None
     link_broken: bool = False
     path: Path | None = None
+    local: Path | None = None
```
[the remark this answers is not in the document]
hexdae: Done, makes a lot of sense

src/atopile/config.py
hexdae: done, you're right. 

In the future we can add the local stuff to ato search paths to avoid copies but it's on optimization

src/atopile/cli/install.py
hexdae: done

hexdae:
Can we run CI on this if it's good to go?

mawildoer (approved):

PR #931, opened 2025-02-04
[RFC] Picker: Implicit `alias_is` for parameters of explicitly-picked parts

sam-mellor:
TODO

### DIFF
--- a/src/faebryk/libs/picker/api/picker_lib.py
+++ b/src/faebryk/libs/picker/api/picker_lib.py
@@ -299,9 +299,23 @@ def get_compatible_parameters(
 
     def _map_param(name: str, param: Parameter) -> tuple[Parameter, P_Set]:
         c_range = component_params.get(name)
-        if c_range is None:
-            c_range = param.domain.unbounded(param)
-        return param, c_range
+        literal = param.try_get_literal()
+
+        match c_range, literal:
+            case None, None:
+                return param, param.domain.unbounded(param)
+            case None, _:
+                return param, literal
+            case _, None:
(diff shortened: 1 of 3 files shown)


src/faebryk/libs/picker/api/picker_lib.py
```diff
@@ -299,9 +299,23 @@ def get_compatible_parameters(
 
     def _map_param(name: str, param: Parameter) -> tuple[Parameter, P_Set]:
         c_range = component_params.get(name)
-        if c_range is None:
-            c_range = param.domain.unbounded(param)
-        return param, c_range
+        literal = param.try_get_literal()
+
+        match c_range, literal:
```
mawildoer, line 304: This might warrant a little explanation. I don't quite understand what's going on here

test/test_regressions.py
```diff
@@ -23,14 +23,18 @@ class BuildError(Exception):
...
 @pytest.mark.slow
 @pytest.mark.regression
 @pytest.mark.parametrize(
-    "repo, env",
+    "repo, env, frozen",
     [
-        ("https://github.com/atopile/spin-servo-drive", {}),
-        ("https://github.com/atopile/esp32-s3", {}),
-        # ("https://github.com/atopile/nonos", {}), # Removing from critical path
+        ("https://github.com/atopile/spin-servo-drive", {}, True),
```
mawildoer, line 32: It'd be enormously valuable for us to only run these once. I wonder if there's a reasonable way to do that, or readily differentiate between the failure modes

PR #947, opened 2025-02-05
add regulator type to fabll lib to define power sink/source

napowderly:
Produces an error when two power supply outputs are connected directly to eachother

### DIFF
--- a/src/faebryk/library/Regulator.py
+++ b/src/faebryk/library/Regulator.py
@@ -1,0 +1,15 @@
+# This file is part of the faebryk project
+# SPDX-License-Identifier: MIT
+
+import faebryk.library._F as F
+from faebryk.core.module import Module
+
+
+class Regulator(Module):
+    power_in: F.ElectricPower
+    power_out: F.ElectricPower
+
+    def __preinit__(self):
+        self.power_out.add(F.Power.is_power_source.impl()())
+        self.power_in.add(F.Power.is_power_sink.impl()())
+        self.power_in.gnd.connect(self.power_out.gnd)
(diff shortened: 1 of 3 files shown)


src/faebryk/library/Regulator.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class Regulator(Module):
+    power_in: F.ElectricPower
+    power_out: F.ElectricPower
+
+    def __preinit__(self):
+        self.power_out.add(F.Power.is_power_source.impl()())
+        self.power_in.add(F.Power.is_power_sink.impl()())
+        self.power_in.gnd.connect(self.power_out.gnd)
```
mawildoer, line 15: Shouldn't be shorted in the base class

src/faebryk/library/regulators.ato
```diff
@@ -22,8 +13,8 @@ module AdjustableRegulator from Regulator:
     # Default value
     i_q = 100uA to 200uA
 
-    assert v_ref * (1 + feedback_div.r_top.resistance / feedback_div.r_bottom.resistance) within v_out
-    assert v_out / (feedback_div.r_top.resistance + feedback_div.r_bottom.resistance) within i_q
+    # assert v_ref * (1 + feedback_div.r_top.resistance / feedback_div.r_bottom.resistance) within v_out
```
mawildoer, line 16: I'm hesitant to remove these, though I know if improves picking speed for you

mawildoer: @iopapamanoglou , FYI, Narayan's stoked that he can get builds back down to 30s by commenting out all the assertions. This is a strong indicator we need it to be that extra 50x faster (at least)

iopapamanoglou: Definitely add those back in the fab ll regulator. You can comment it out locally if you need the extra speed for now.

iopapamanoglou (dismissed):

src/faebryk/library/regulators.ato
mawildoer: ** where they were in ato is good, not the base fabll regulator

mawildoer (approved):

PR #954, opened 2025-02-06
`PickError` improvement

mawildoer:
- Raises an exception group of `PickError`s in the case of a server failure
- Adds `iter_leaf_exceptions` as a helper to flatten `ExceptionGroup`s
- Adds params to error output

Screenshot 2025-02-07 at 10 19 55


mawildoer:
Screenshot 2025-02-07 at 11 14 31

PR #523, opened 2024-11-16
faebryk: Core: Better Mifs split path algo

iopapamanoglou:
# faebryk: Core: Improve MIFS split path algo

# Description

Current split paths are slow and don't 100% work.
This algo does some smart pruning.
It only looks at a single path in a split at a time.

Current state: 
- Cuts down lots of paths in connected case (see hull i2c)
- Doesn't seem to make a diff in no connect
- But that's not a problem (we just want quicker discovery of actual paths so the heuristic limit can be reduced)
- one open point: pathfinder.cpp ~450: valid but not complete paths should hibernate and be woken up, this can be done in a seperate PR tho
- [ ] Tests all pass
- [ ] cellsim works

image

Demo:
```python
Path(3)[
    *B948.self
    *B948.children
    *B948.lower2.parent]
New split
Split: Path(3)[
    *B948.self
    *B948.children
    *B948.lower1.parent]
Not waiting
Handle valid split branch: Path(10)[
    *B948.self
    *B948.children
    *B948.lower2.parent
    *B948.lower2.self
    *B948.lower2.connected
    *BA18.lower2.connected
    *BA18.lower2.self
    *BA18.lower2.parent
    *BA18.children
    *BA18.self]
Check complete branch for Path(2)[
    *B948.self
    *B948.children]
Wake up path: Path(3)[
    *B948.self
    *B948.children
    *B948.lower1.parent]
Wake signal, checking: 1 paths
Reschedule awoken path: Path(3)[
    *B948.self
    *B948.children
    *B948.lower1.parent]
Handle valid split branch: Path(10)[
    *B948.self
    *B948.children
    *B948.lower1.parent
    *B948.lower1.self
    *B948.lower1.connected
    *BA18.lower1.connected
    *BA18.lower1.self
    *BA18.lower1.parent
    *BA18.children
    *BA18.self]
Check complete branch for Path(2)[
    *B948.self
    *B948.children]
Complete branch found
All branches complete
```

`./test/runpytest.sh -k test_mif_connect_hull`
```
# Before
┏━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━┳━━━━━━━━━┳━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━┓
┃ func                 ┃   in ┃ weak in ┃  out ┃   filt ┃ weaker ┃ stronger ┃    time ┃ time/in ┃
[...]

PR #963, opened 2025-02-10
More robust net name conflict checking

mawildoer:
Fixes: #984


test/core/test_mif_connect.py
```diff
@@ -138,6 +138,31 @@ def test_chains_mixed_shallow_nested():
     assert el[0].is_connected_to(el[2])
 
 
+def test_loooooong_chain():
```
mawildoer, line 141: Likely fixed by https://github.com/atopile/atopile/pull/523

PR #964, opened 2025-02-10
add larger common package types

napowderly:
Add common package types above 0805


mawildoer (changes requested):
Needs the matching footprints and updates to `src/atopile/package/__init__.py` too

napowderly:
> Needs the matching footprints and updates to `src/atopile/package/__init__.py` too

Thanks, updated __init__ and copied in packages from Kicad lib

PR #969, opened 2025-02-11
Profile build times for regressions

mawildoer:


mawildoer:
Bypassing pytest failing since fix is on main

PR #970, opened 2025-02-11
Remove unused dependencies

mawildoer:


src/atopile/cli/create.py
```diff
@@ -458,54 +457,6 @@ def add_build_target(config_data: dict, new_data: dict):
     )
 
 
-@create_app.command(hidden=True)
```
mawildoer, line 461: Replaced in #960

PR #974, opened 2025-02-12
Buildutil: Enable skipping solving via config flag

sam-mellor:
Allows skipping solving with a config flag (unsafe). Recommended for development use only.

### DIFF
--- a/src/faebryk/core/solver/nullsolver.py
+++ b/src/faebryk/core/solver/nullsolver.py
@@ -6,6 +6,41 @@ from faebryk.core.solver.solver import Solver
 from faebryk.libs.sets.sets import P_Set
 
 
+class SuperSuperSet(P_Set):
+    """Is a superset of anything"""
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+
+    def is_superset_of(self, other: P_Set) -> bool:
+        return True
+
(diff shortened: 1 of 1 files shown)


src/faebryk/core/solver/nullsolver.py
```diff
@@ -0,0 +1,77 @@
+from typing import Any
+
+from faebryk.core.cpp import Graph
+from faebryk.core.parameter import Expression, Parameter, Predicate
+from faebryk.core.solver.solver import Solver
+from faebryk.libs.sets.sets import P_Set
+
+
+class SuperSuperSet(P_Set):
```
mawildoer, line 9:
```suggestion
class SuperDuperSet(P_Set):
```

src/atopile/buildutil.py
```diff
@@ -59,15 +61,25 @@
     C_kicad_pcb_file,
 )
 from faebryk.libs.picker.picker import PickError, pick_part_recursively
-from faebryk.libs.util import KeyErrorAmbiguous
+from faebryk.libs.util import ConfigFlag, KeyErrorAmbiguous
 
 logger = logging.getLogger(__name__)
 
+SKIP_SOLVING = ConfigFlag("SKIP_SOLVING", default=False)
```
mawildoer, line 68: I reckon we might need to turn this on by default

mawildoer (approved):

PR #977, opened 2025-02-14
updated .value to .resistance

napowderly:
.value no longer supported in std library parts, need to use .resistance instead


mawildoer:
This should also raise an exception

mawildoer (approved):

PR #991, opened 2025-02-19
Upgrade `pytest-history` version

mawildoer:
We should publish this properly, so we can better manage versions. This will have to do for the second


mawildoer:
Skipping review

PR #992, opened 2025-02-19
More cores for pytest

mawildoer:


mawildoer:
No appreciable difference, or worse

PR #998, opened 2025-02-20
Allow pytest from trusted forks

mawildoer:


mawildoer:
Tested with https://github.com/atopile/atopile/pull/999

PR #1001, opened 2025-02-20
Don't deploy docker container on forks

mawildoer:


.github/workflows/deploy.yml
```diff
@@ -170,15 +170,24 @@ jobs:
...
             type=sha
 
+      - id: PUSH_CONTAINER
+        run: >-
+          echo "PUSH_CONTAINER=${{
+            ( github.event_name == 'pull_request' &&
+            github.event.pull_request.head.repo.full_name == github.repository ) ||
+            github.event_name == 'pull_request'
+          }}" >> $GITHUB_OUTPUT
+
```
iopapamanoglou, line 188: thats not right (prohibits non-pr runs)

Should be:
- dont run this is foreign repos (except if PR target is our repo)

Issue #1008, opened 2025-02-23
PCB transformer function to hide all designators


JasperZP:
### Feature Request

Designator visibility is not preserved, so hiding designators is a manual task every time the PCB is generated.
Related: #868 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1013, opened 2025-02-24
Faebryk: exporters: PCB: KiCad: Add `hide_all_designators()` method

jrhrsmit:
# Faebryk: exporters: PCB: KiCad: Add `hide_all_designators()` method

# Description

Add functionality in the KiCad PCB transformer to hide all designators.
Fixes https://github.com/atopile/atopile/issues/1008

# Checklist

### DIFF
--- a/src/faebryk/exporters/pcb/kicad/transformer.py
+++ b/src/faebryk/exporters/pcb/kicad/transformer.py
@@ -1303,6 +1303,15 @@ class PCB_Transformer:
         LEFT = auto()
         RIGHT = auto()
 
+    def hide_all_designators(
+        self,
+    ) -> None:
+        for _, fp in self.get_all_footprints():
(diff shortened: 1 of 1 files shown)


src/faebryk/exporters/pcb/kicad/transformer.py
```diff
@@ -1303,6 +1303,15 @@ class Side(Enum):
         LEFT = auto()
         RIGHT = auto()
 
+    def hide_all_designators(
```
mawildoer, line 1306: Definitely makes sense! Is this behaving well? We've had some issues with KiCAD ignoring this field

jrhrsmit: Yes, the trick is to not only set the `hide` field of the reference property of the footprint itself, but also the `effects.hide` field of the reference text object

PR #1016, opened 2025-02-27
Core: Solver: Fix no supersets for package requirements

sam-mellor:


iopapamanoglou:
This PR will def not work with #1012

PR #1029, opened 2025-03-04
Picker: Check for unspecified parameters

sam-mellor:
Adds pre-pick verification that all parameter data is present for parts which are explicitly specified (i.e. by manufacturer/part number or by supplier ID). Enables earlier and more comprehensible build failures, rather than letting picking fail due to solver contradictions.

### DIFF
--- a/src/faebryk/core/solver/symbolic/structural.py
+++ b/src/faebryk/core/solver/symbolic/structural.py
@@ -191,7 +191,7 @@ def remove_congruent_expressions(mutator: Mutator):
         mutator._mutate(expr, repres[eq_id])
 
 
-@algorithm("Alias classes")
+@algorithm("Alias classes", terminal=False)
 def resolve_alias_classes(mutator: Mutator):
     """
     Resolve alias classes


src/faebryk/libs/app/checks.py
```diff
@@ -2,10 +2,103 @@
...
+def check_parameters(parameters: set[Parameter], G: Graph, solver: Solver):
+    """
+    Basic checks to find issues prior to picking.
+
+    - For parts which are explicitly specified, all parameters constrained to something
+      other than their domain must be specified
+    """
+    logger.info(f"Checking {len(parameters)} parameters")
+
+    solver = cast_assert(DefaultSolver, solver)  # FIXME
```
iopapamanoglou, line 64: need to support nullsolver

src/faebryk/libs/app/checks.py
```diff
@@ -2,10 +2,103 @@
...
+    with accumulate(ParameterError) as accumulator:
+        for module in explicit_picks:
+            assert isinstance(module, Module)
+
+            for orig_param in module.get_parameters():
+                with accumulator.collect():
+                    if (
+                        maps_to := mutation_map.map_forward(orig_param).maps_to
+                    ) is None:
+                        continue
```
iopapamanoglou, line 84: this should never happen

src/faebryk/libs/app/checks.py
```diff
@@ -2,10 +2,103 @@
...
+
+                    param = cast_assert(Parameter, maps_to)
+                    preds = param.get_operations(Predicate, constrained_only=True)
```
iopapamanoglou, line 87:
```suggestion
                    preds = param.get_operations(ConstrainableExpression, constrained_only=True)
```

iopapamanoglou: recursive=True

PR #1038, opened 2025-03-05
Picker: Better error messages for contradictions and incompatibility

sam-mellor:
Improves error reporting for contradictions and incompatible components.


iopapamanoglou (changes requested):
why replace the indented_contained  everywhere?

iopapamanoglou (approved):

PR #1048, opened 2025-03-06
add setter/getter for suggested_net_name in ato

napowderly:
provides a way to softly set a net name in ato

Future upgrade:
- Prioritize suggested name based on project hierarchy (see below)
- Prefix suggested names with source id when there are conflicts, eg cell_1_gnd instead of gnd_8


napowderly:
I couldnt find a way to prioritize suggested names that are higher up in the project structure.

Eg:

In the case below, we would always want to give the name "apple", but we dont currently do so.

module App:
    sub_module = new SubModule
    sub_module.test_signal.suggest_net_name = "apple"

module SubModule:
    signal test_signal
    test_signal.suggest_net_name = "banana"

PR #1054, opened 2025-03-07
Libs: Launch `pcbnew` outside Python venv

sam-mellor:


napowderly (approved):
Tested on Mac, doesnt seem to have broken anything with the plugin

PR #1061, opened 2025-03-10
Ato: Better errors for invalid package assignment

sam-mellor:
Un-prefixes list of valid packages in error message, to match desired input, and adds tracebacks.

Before:
image

After:
image


iopapamanoglou (dismissed):
I don't understand what's going on here.
Why is the package name 0402 and not C0402?
We need to be consistent, in F.has_package we differentiate between C,L,R.
If that is not the way we want to go we should get rid of that.
Else we should forbid the other way around.

napowderly (approved):

Issue #773, opened 2025-01-02
`ato build -t all` exports files to wrong folder


ruben-iteng:
### Current Behaviour

Currently all export files are put into the `build` folder.

### Expected Behaviour

I would expect a better folder structure. 
e.g. `build/elec/production/`

### Version

0.3.0-dev

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1071, opened 2025-03-11
Split build targets into their own directory

napowderly:
Much cleaner output of build artifacts.
Screenshot 2025-03-11 at 3 33 08 PM


napowderly:
addresses #773

PR #1121, opened 2025-03-20
Generate all layers

mawildoer:
See https://docs.kicad.org/9.0/en/cli/cli.html


napowderly (approved):
Tested with a 4 layer board locally, works great

PR #1126, opened 2025-03-24
Exporters: PCB: Fix missing support for arcs in polygons

sam-mellor:
Fixes pcbnew crash because we export an empty `zone.polygon.pts`.


napowderly (approved):
Works for me! Thanks mate

Issue #1128, opened 2025-03-26
Multiple specializations (`->`) cause `Trait Not Found` error


mawildoer:
### Current Behaviour

Approximately this:
```ato
a = new Thing
a -> Other
a -> Another
```

Causes this:
```
[17:09:19] ERROR    Trait Not Found                                                                                                                                                                                                                                                                            
                    Trait PCB_Transformer.has_linked_kicad_pad not found in Pad[*A6A8.capacitor_can.footprint.pins[1]]                                                                                                                                                                                         
```

### Expected Behaviour

The program exited with all checks "ok"

### Possible Solution

_No response_

### Version

c381e92b28bd6744dee0f91c8eaf24ce4783c93b

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1129, opened 2025-03-26
Fix: Buildutil: Use most recent Graph

iopapamanoglou:
G becomes invalid when modified, need to update the reference.

Fixes `Trait has_designator not found` errors.


mawildoer (approved):

mawildoer:
Likely also fixes #1128

PR #1141, opened 2025-04-01
Fix: Save & Load params

iopapamanoglou:
Dump parameter aliases of picked modules to pcb attributes, so when building with `--keep-picked-parts` we can load the values for verification.


napowderly (approved):
Built all HIL build targets with a clean build then --keep-picked-parts, no issues

PR #1144, opened 2025-04-02
Add Posthog Telemetry

mawildoer:
Duplicate (for now) telemetry over to Posthog

### DIFF
--- a/src/atopile/telemetry.py
+++ b/src/atopile/telemetry.py
@@ -15,14 +15,20 @@ import hashlib
 import logging
 import subprocess
 import time
+from contextlib import contextmanager
 from pathlib import Path
 from typing import Optional
 
+import posthog
 import requests
 from attrs import asdict, define, field
 from ruamel.yaml import YAML
 
 log = logging.getLogger(__name__)
+
+# Public API key, as it'd be embedded in a frontend
+posthog.api_key = "phc_IIl9Bip0fvyIzQFaOAubMYYM2aNZcn26Y784HcTeMVt"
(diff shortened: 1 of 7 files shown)


src/atopile/telemetry.py
```diff
@@ -15,15 +15,21 @@
...
 
+import posthog
 import requests
 from attrs import asdict, define, field
 from ruamel.yaml import YAML
 
 log = logging.getLogger(__name__)
 
+# Public API key, as it'd be embedded in a frontend
+posthog.api_key = "phc_IIl9Bip0fvyIzQFaOAubMYYM2aNZcn26Y784HcTeMVt"
```
mawildoer, line 30: I'm like ~60% sure this is safe. @iopapamanoglou

src/atopile/telemetry.py
```diff
@@ -15,15 +15,21 @@
...
+posthog.host = "us.i.posthog.com"
```
mawildoer, line 31: It'd be nice to reverse-proxy this to atopileapi.com

src/atopile/telemetry.py
```diff
@@ -183,3 +191,30 @@ def get_project_id() -> Optional[str]:
 
     except subprocess.CalledProcessError:
         return None
+
+
+@contextmanager
+def log_to_posthog(event: str, properties: dict | None = None):
```
mawildoer, line 197: Made this helper because in call cases, I wanted the project metadata and duration, which is otherwise a little annoying to capture. Works neatly for exceptions too!

PR #1181, opened 2025-04-18
Fix: Test: Don't xfail hil

iopapamanoglou:
Also disable profiling

### DIFF
--- a/test/test_regressions.py
+++ b/test/test_regressions.py
@@ -25,11 +25,13 @@ class BuildError(Exception):
     """Failed to build the project."""
 
 
+ENABLE_PROFILING = False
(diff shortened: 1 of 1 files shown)


mawildoer:
@iopapamanoglou this warning makes me think the regression tests aren't being run: https://github.com/atopile/atopile/actions/runs/14542576346/job/40803200956?pr=1181#step:11:83

Are we confident it's all working? If so, perhaps it's because `pytest` collects any classes prefixed with "Test", and so it's attempting to collect this infra class (a leading underscore or rename would fix the warning)

mawildoer (approved):

PR #1188, opened 2025-04-23
Fix frozen file comparison

napowderly:
Frozen check was comparing the wrong thing.


napowderly:
I didnt really like the alternative error message any better, so kept as is. @iopapamanoglou

iopapamanoglou (approved):

PR #1190, opened 2025-04-24
Libs: Parts: Prototype Lifecycle management & 3D Models

iopapamanoglou:
- Enable/fix 3D models
- Introduce Part lifecycle management
- handle symbols
- (partial) embedded file (encode/decode only) support


iopapamanoglou:
Superseded by #1196

PR #1210, opened 2025-04-30
Fix: Run solver even if only explicit picks

iopapamanoglou:
Lines that backed up tree were too late.


napowderly:
Screenshot 2025-04-30 at 4 34 02 PM
Fixed my addressor issue!

napowderly (approved):

PR #1257, opened 2025-05-19
Fix: ato: Lock grammar compile

iopapamanoglou:
- Use global lock for antlr compile


iopapamanoglou:
TODO
- [ ] add antrl.lock to gitignore

PR #1270, opened 2025-06-13
Fix: STEP parser aborts on invalid utf-8 in header

### LINKED ISSUE #1266: Part lifecycle manager: Some step files are not utf-8 encoded
### Current Behaviour

utf-8 decoding is not possible with some step files 
https://github.com/atopile/atopile/blob/main/src/faebryk/libs/kicad/fileformats_latest.py#L2153

### Expected Behaviour

Step files are converted to utf-8 or error is gracefully handled.

### Possible Solution

#pr number

### Version

0.9.0

### Relevant log output

```shell
tuf-8 codec can’t decode byte 0xc8 in position 160: invalid continuation byte
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

tachiniererin:
# Replace invalid UTF-8 characters when decoding STEP files

## Description

This patch instructs the utf-8 decoder to replace any invalid characters according to https://docs.python.org/3/library/codecs.html#error-handlers.
I'm not 100% sure this is the best approach, e.g. one option would be to try different code-pages to decode it into something meaningful, but also not sure how relevant/useful this would be for the STEP file header, KiCad parses it just fine after replacement and the user first and foremost expects this operation not to fail due to some slightly malformed data.

An example where this previously failed to generate the part is C2060364.

Fixes #1266

## Checklist

### DIFF
--- a/src/faebryk/libs/kicad/fileformats_latest.py
+++ b/src/faebryk/libs/kicad/fileformats_latest.py
@@ -2149,8 +2149,10 @@ class C_kicad_model_file:
     def header(self) -> str:
         # Extract header section between HEADER; and ENDSEC; using regex
 
-        # Read till DATA; token
-        non_data = first(lazy_split(self._raw, b"DATA;")).decode("utf-8")
+        # Read till DATA; token, replace any invalid UTF-8 characters that may occur
+        non_data = first(lazy_split(self._raw, b"DATA;")).decode(
+            "utf-8", errors="replace"
+        )
 
         pattern = r"HEADER;(.*?)ENDSEC;"
         match = re.search(pattern, non_data, re.DOTALL)


iopapamanoglou:
Thanks for the PR!
@ruben-iteng found that apparently the step files are (sometimes?) encoded in latin-1

iopapamanoglou:
> Ok, that means it could be recoverable. It should be possible to try and decode the sequence that is failing with a decode error handler then.

I think STEP files are supposed to be always encoded in latin-1.
So we don't need to try multiple codecs .
Having the errors='replace' is still a nice addition, because we don't really do anything important with the step anyway. We just extract an internal field for naming.

tachiniererin:
The encoding was GB 18030 in the end, so not very conformant :D

As for the step format, ISO 10303-21:2016-03, section 5.2:
>  NOTE     The set of LATIN_CODEPOINT character is equivalent to the basic alphabet in the first and second editions of ISO 10303-21. The UTF-8 representation of code points U+0020 to U+007E is the same as the ISO/IEC 8859-1 characters G(02/00) to G(07/14) that defined the basic alphabet in earlier editions. Use of HIGH_CODEPOINT characters within the exchange structure can be avoided when compatibility with previous editions of ISO 10303-21 is desired.

So from the 2016 edition on UTF-8 is ok, but codepoints should be ignored. Should i keep the replace still or change it to ignore? The possibility still exists that might break things somewhere else if the codepoints don't get replaced.

iopapamanoglou:
Thanks for the PR, resumed it in #1283, so we can run the CI suite.

Issue #1279, opened 2025-06-17
Create part fails when trying to fetch a footprint without a 3d model


tachiniererin:
### Current Behaviour

`UserException: 'NoneType' object has no attribute 'name' ` in src/libs/faebryk/picker/lcsc.py:375 with part C22427496.

### Expected Behaviour

Create the part without a 3d model.

### Possible Solution

#1280

### Version

0.9.1

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #1265, opened 2025-06-08
Part lifecycle manager: 3D file is not included


ruben-iteng:
### Current Behaviour

When no build dir and no part dir is present, then building your project (and thus downloading/creating the part lib etc, the generated parts will not include the step file (in parts/<partname>/) but the step file is in the build/cache/parts/<lcscid>/ folder.

Reproduce by:
1. deleting build and parts folder
2. build project
3. …no 3D files in pars folder, but they are in the build/cache/.. folder
4. delete parts folder
5. build project
6. now there are 3D files in the parts folder

### Expected Behaviour

3D files are included in the parts folder

### Possible Solution

-

### Version

0.9.0

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

iopapamanoglou:
Slightly related:

When making PCB for first time and exporting using kicad cli, no 3d models.
Saving pcb with pcbnew and rerunning works

iopapamanoglou:
Fixed in #1283

PR #1283, opened 2025-06-17
Fix: Parts: Multiple model issues

iopapamanoglou:
Fixes multiple issues with models in the parts lifecycle.

Fixes:
#1279: Handle missing 3dmodel in easyeda
#1266: STEP file decoding utf-8
#1265: No 3Dmodels on first run with fresh build folder

Based on:
#1280
#1270 

Thanks to @ruben-iteng and @tachiniererin for the reporting and fixes.


src/faebryk/libs/picker/lcsc.py
```diff
@@ -212,11 +211,15 @@ def load(cls, path: Path):
         return out
 
     @classmethod
-    def from_api(cls, footprint: ee_footprint, model_path: str):
+    def from_api(cls, footprint: ee_footprint, model_path: str | None):
         exporter = ExporterFootprintKicad(footprint)
         _fix_3d_model_offsets(exporter)
+        if model_path is None:
```
sam-mellor, line 217: I'd rather have this check in the exporter

iopapamanoglou: this is not our code

PR #1316, opened 2025-06-26
VSCE: Add 3D model viewer

sam-mellor:
Adds PCBA 3D model previewing to the VSCode extension.

image


iopapamanoglou (changes requested):
- [ ] missing icon
image

- [x] two windows bugs (I'll fix them)
- [ ] hint to user 'build your target with -t mfg-data`

image

PR #1317, opened 2025-06-26
Buildutil: Split out GLB export into new `3d-model` target

sam-mellor:
Enabled by default because it's used for the VSCode extension's 3D model preview. Rejected just making `mfg-data` default-enabled for performance.


iopapamanoglou (changes requested):
Depends on kicad being installed.
Introduce soft-default targets that depend on kicad.
If kicad is not installed, just spit out warning

iopapamanoglou (approved):

PR #1326, opened 2025-06-30
FIX: Connect DifferentialPair References 

nickkrstevski:
Differential pair signals should be referenced to a shared reference. Common mode voltage would be expected to be wrt to the same reference.

Added single_electric_reference trait to connect all references.


src/faebryk/library/DifferentialPair.py
```diff
@@ -20,6 +20,12 @@ class DifferentialPair(ModuleInterface):
         tolerance_guess=10 * P.percent,
     )
 
+    @L.rt_field
+    def single_electric_reference(self):
+        return F.has_single_electric_reference_defined(
+            F.ElectricLogic.connect_all_module_references(self)
```
iopapamanoglou:
```suggestion
            F.ElectricSignal.connect_all_module_references(self)
```

iopapamanoglou (changes requested):
Run pre-commit and for the rest LGTM

iopapamanoglou:
Merged in #1331

PR #1334, opened 2025-07-02
VSCE: Auto rebuild 3D model

iopapamanoglou:
- rebuild glb on pcb change
- fix filewatcher
- disable 3d as default build target

## TODO
- [ ] check & fix on macos & windows


src/vscode-atopile/src/common/3dmodel.ts
```diff
@@ -46,10 +46,51 @@ export function eqThreeDModel(a: ThreeDModel | undefined, b: ThreeDModel | undef
...
+    if (!pcb || !pcb.exists) {
+        return;
+    }
+
+    const model = getThreeDModelForBuild();
+    if (!model) {
+        return;
+    }
+    // check model newer than pcb
+    if (model && model.exists && fs.statSync(model.path).mtimeMs > fs.statSync(pcb.path).mtimeMs) {
```
sam-mellor:
```suggestion
    if (model && model.exists && fs.statSync(model.path).mtimeMs >= fs.statSync(pcb.path).mtimeMs) {
```

Fail safe if `mtime` isn't being updated.

PR #1343, opened 2025-07-04
Fix: Net names determinism

iopapamanoglou:
If mif connected without parent was still undeterministic.
- also fixed '-0' not being appended in duplicates

### DIFF
--- a/src/faebryk/exporters/netlist/graph.py
+++ b/src/faebryk/exporters/netlist/graph.py
@@ -208,12 +208,14 @@ def attach_net_names(nets: Iterable[F.Net]) -> None:
     # sort nets by
     # 1. name of first connected interface (remove hex)
     # 2. number of connected interfaces
+
+    def stable_node_name(m: ModuleInterface) -> str:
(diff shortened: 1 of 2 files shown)


src/faebryk/exporters/netlist/graph.py
```diff
@@ -203,6 +205,17 @@ def attach_net_names(nets: Iterable[F.Net]) -> None:
...
         if not n.has_trait(F.has_overriden_name)
     }
+    # sort nets by
+    # 1. name of first connected interface (remove hex)
+    # 2. number of connected interfaces
+    unnamed_nets = dict(
+        sorted(
+            unnamed_nets.items(),
+            key=lambda it: [
+                re.sub(r"^\*[0-9A-F]+\.", "", m.get_full_name()) for m in it[1]
```
sam-mellor: This seems fragile

sam-mellor: I'd rather have a `get_full_name()` parameter to leave out the hex

PR #1359, opened 2025-07-15
SEXP: Add kicad dru file format support

ruben-iteng:
### DRU file format support
- Add KiCad DRU (Design Rule) file format in Zig with full encode/decode based on the [official KiCad source](https://gitlab.com/kicad/code/kicad/-/tree/master/pcbnew/drc)
- Add `loadsFlat`/`dumpsFlat` to the sexp framework for top-level unwrapped S-expressions (DRU files don't have a wrapping symbol like other KiCad formats)
- Add custom `encode` method support in `structure.zig` (mirroring existing custom `decode`) — needed for `ValueWithUnit` which concatenates value and unit into a single token (e.g. `0.4mm`) and for `Constraint` which has variant-dependent structure based on constraint type

### Additional fixes
- Add proper enum support in the Python API (typed enums instead of raw `int`/`str`)
- Add default values for optional fields in `.pyi` stubs (fixes linting)
- Filter internal types (SExp, SExpValue) from generated Python stubs
- Fix float rounding bug (`10e6` → `1e6` = 6 decimal places, not 7)
- Simplify string escape/unescape in `ast.zig` (KiCad only escapes quotes)

## Motivation
Reading and writing DRU files is essential for making KiCad DRC useful. With this we can create and auto-apply PCB manufacturer-specific DRC rules alongside generic constraints like differential pair gaps or power rail trace widths.


ruben-iteng:
ported

PR #1367, opened 2025-07-16
Libs: Parts: Ignore datasheet removals for part definition updates

sam-mellor:
Preserves datasheet in part definition if not returned from EasyEDA, which happens often. Fixes issue with unstable artifacts.


src/faebryk/libs/ato_part.py
```diff
@@ -103,7 +103,21 @@ def __post_init__(self):
...
+            did have one previously, filter that update out of the diff so we don't
+            then remove it from the part definition.
+            """
+
+            if (datasheet := diff.get(".datasheet")) is not None:
+                before, after = datasheet
+                if before is not None and after is None:
+                    del diff[".datasheet"]
+
+            return diff
```
iopapamanoglou: Move t https://github.com/atopile/atopile/blob/e69881889c97f3146ebcbc150225119d8ff2103a/src/faebryk/libs/part_lifecycle.py#L302

iopapamanoglou (approved):

PR #1387, opened 2025-07-26
CLI: Fix legacy config dir cleanup

sam-mellor:
Cleanup all known legacy config dirs, with safety check for unexpected dir content

### DIFF
--- a/src/faebryk/libs/paths.py
+++ b/src/faebryk/libs/paths.py
@@ -13,6 +13,25 @@ from faebryk.libs.util import once
 
 @once
 def get_config_dir() -> Path:
+    def _cleanup_legacy_dir(legacy_path: Path, out: Path) -> None:
+        known_files = [
+            "config.yaml",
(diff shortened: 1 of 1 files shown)


iopapamanoglou (approved):

src/faebryk/libs/paths.py
```diff
@@ -26,14 +26,23 @@ def get_config_dir() -> Path:
...
+            "config.yaml",
+            "configured_for.yaml",
+            "telemetry.yaml",
+        ]
+
+        # chronological order
+        for legacy_path in [Path.home() / ".atopile", Path.home() / "atopile"]:
+            if legacy_path.exists() and all(
+                file in known_files for file in legacy_path.iterdir()
+            ):
```
iopapamanoglou: always move/remove the known files
rm the dir if else empty

iopapamanoglou (changes requested):

PR #1399, opened 2025-07-29
ato: Support enums in template args

sam-mellor:
Adds support for providing enum values via template args.

```ato
#pragma experiment("TRAITS")
#pragma experiment("MODULE_TEMPLATING")

import Resistor
import has_package_requirements

module Resistor0805 from Resistor:
    trait has_package_requirements<size="I0805">

module App:
    r = new Resistor0805
```

### DIFF
--- a/test/front_end/test_front_end.py
+++ b/test/front_end/test_front_end.py
@@ -1934,7 +1934,11 @@ class ModuleWithTypingUnionEnumOrNone(L.Module):
     [
         # Basic enum types
         ("ModuleWithIntEnum", "<value=2>", {"_value": _IntEnumForTests.VALUE_2}),
-        ("ModuleWithStrEnum", '<mode="OPTION_B">', {"_mode": _StrEnumForTests.OPTION_B}),
+        (
+            "ModuleWithStrEnum",
+            '<mode="OPTION_B">',
+            {"_mode": _StrEnumForTests.OPTION_B},
+        ),
         # Optional enums with modern syntax
         ("ModuleWithOptionalEnum", "", {"_color": None}),
         ("ModuleWithOptionalEnum", '<color="RED">', {"_color": F.LED.Color.RED}),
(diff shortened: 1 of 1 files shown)


src/faebryk/core/node.py
```diff
@@ -493,6 +493,7 @@ def _setup(self, *args, **kwargs) -> None:
                     if f_name in base.__dict__:
                         f = getattr(base, f_name)
                         f(self)
+
```
sam-mellor:
```suggestion
```

src/atopile/front_end.py
```diff
@@ -1492,7 +1539,31 @@ def _new_node(
...
+                    )
+
+                if isinstance(arg_value, str):
+                    try:
+                        kwargs[arg_name] = _try_upgrade_str_to_enum(
+                            type_hints, arg_name, arg_value
+                        )
+                    except _EnumUpgradeError as ex:
+                        raise errors.UserInvalidValueError.from_ctx(
+                            origin=None,  # TODO: add context
```
sam-mellor, line 1561: This is bad but seems tricky to fix and I'm assuming the front-end updates will make it irrelevant

test/front_end/test_front_end.py
```diff
@@ -1727,3 +1727,113 @@ def test_assign_to_enum_param(bob: Bob):
...
+        import ResistorWithSize
+
+        module App:
+            r = new ResistorWithSize<size="<invalid size>">
+        """
+    )
+
+    tree = parse_text_as_file(text)
+    with pytest.raises(errors.UserInvalidValueError):
+        bob.build_ast(tree, TypeRef(["App"]))
```
iopapamanoglou, line 1839: More tests:
- IntEnum
- StrEnum
- arg: Optional[Enum]
- arg: Enum1 | Enum2 | Enum3
- arg: str | Enum
- arg: str | StrEnum
- arg: Enum | None

src/atopile/front_end.py
iopapamanoglou: Yeah no worries, the type-first recipe stuff will make this trivial

iopapamanoglou (changes requested):
Looks good, just add some extra tests (of which I think some are going to fail).

PR #1403, opened 2025-07-30
Buildutil: More robust PCB comparison

sam-mellor:
Makes PCB comparison for `--frozen` check more robust to changes that don't affect layout. Diff is displayed inline to aid debugging.


src/faebryk/libs/sexp/dataclass_sexp.py
```diff
@@ -674,3 +675,41 @@ def _set(k, v, it):
...
+            for f in fields(obj):
+                if f.name in field_names:
+                    setattr(obj, f.name, None)
+                else:
+                    setattr(
+                        obj, f.name, _filter_fields(getattr(obj, f.name), field_names)
+                    )
+            return obj
+        case _:
+            return obj
```
iopapamanoglou: might make sense to extend _iterate_tree (setter return value for filtering)

src/faebryk/libs/sexp/dataclass_sexp.py
```diff
@@ -674,3 +675,41 @@ def _set(k, v, it):
...
+    Returns a copy of a dataclass object (or collection of dataclass instances),
+    with all properties listed in field_names recursively removed.
+    """
+    out = deepcopy(obj)
+    _filter_fields(out, field_names)
+    return out
+
+
+def filter_uuids[T: DataclassInstance](obj: T) -> T:
+    return filter_fields(obj, field_names=["uuid"])
```
iopapamanoglou: this is not dataclass_sexp generic, very much a kicad thing

src/faebryk/libs/util.py
```diff
@@ -2383,10 +2427,12 @@ def compare_dataclasses[T](
...
         return {"before": b, "after": a}
 
-    if type(before) is not type(after) and not all(
-        isinstance(x, (int, float)) for x in (before, after)
-    ):
-        return {"": (before, after)}
+    if (
+        type(before) is not type(after)
+        # dataclasses equality is independent of type
+        and not (is_dataclass(before) and is_dataclass(after))
```
iopapamanoglou: the comparison in 2455 relies on them being the same type

src/atopile/buildutil.py
```diff
@@ -77,6 +86,71 @@ def _get_solver() -> Solver:
...
+                        ],
+                        headers=["Path", "Before", "After"],
+                    ),
+                ),
+                title="Frozen failed",
+                # No markdown=False here because we have both a command and paths
+            )
+        else:
+            logger.info("No changes to layout. Passed --frozen check.")
+    elif original_pcb_file == pcb_file:
```
iopapamanoglou, line 148: why compare again, we did already in 92

sam-mellor: Different comparisons for different purposes — if there are non-meaningful changes, we want `frozen` to pass but to still update the file.

PR #1404, opened 2025-07-30
Examples: Add led badge and minimal ESP32 example

napowderly:
Add example of esp32 minimal design and led badge that utilize packages.


napowderly:
> See #1398

done

PR #1407, opened 2025-07-30
CLI: Make faster

iopapamanoglou:
- Remove module level slow imports in cli


src/faebryk/exporters/pcb/kicad/transformer.py
```diff
@@ -9,14 +9,24 @@
...
+    TypeVar,
+)
 
 import numpy as np
 from deprecated import deprecated
 from more_itertools import first
 from shapely import Polygon
 
-import faebryk.library._F as F
+# import faebryk.library._F as F
```
sam-mellor, line 29:
```suggestion
```

PR #1410, opened 2025-07-31
KiPlugin: Remove layout manifest

iopapamanoglou:
Removes need to compile for layout reuse

- encode nested addresses into pcb
- simplify logic & use graph
- handle weird nested & duplicate entries

# TODO
- [x] Needs some extra testing
- [x] Submodules
- [x] fix drc thing

### DIFF
--- a/src/faebryk/library/PCB.py
+++ b/src/faebryk/library/PCB.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, override
 import faebryk.library._F as F
 from faebryk.core.module import Module
 from faebryk.core.node import Node
+from faebryk.core.reference import reference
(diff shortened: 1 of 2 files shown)


src/faebryk/library/PCB.py
```diff
@@ -132,17 +130,32 @@ def __check_post_pcb__(self):
                     pcb, shorts, not_connected, drc_report.coordinate_units
                 )
 
+    # TODO use reference
     class has_pcb(Module.TraitT.decless()):
-        class has_pcb_ref(F.has_reference.decless()):
```
sam-mellor, line 136: This breaks DRCs

PR #1415, opened 2025-08-02
Fix: footprint flipping issue in layout reuse plugin

napowderly:
When moving footprints from top to bottom layer, the plugin now correctly mirrors the footprint geometry about the Y-axis. This fix includes:

- Mirror pad positions and rotations about Y-axis
- Mirror text and property positions about Y-axis
- Mirror geometry objects (lines, arcs) about Y-axis
- Properly handle circle center points to prevent distortion

This ensures footprints maintain correct orientation when viewed from the top of the board after being moved to the bottom layer.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/faebryk/exporters/pcb/kicad/transformer.py
+++ b/src/faebryk/exporters/pcb/kicad/transformer.py
@@ -1189,34 +1189,37 @@ class PCB_Transformer:
 
             fp.layer = _flip(fp.layer)
 
-            # Mirror the footprint geometry about y-axis when flipping layers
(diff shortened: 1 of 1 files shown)


napowderly:
Screenshot 2025-08-04 at 11 15 27 AM
Only used in layout sync and component positioning. For component positioning we dont currently do any placement where the component is flipped, but if we did it would be broken, so this should fix that for future use too.

iopapamanoglou (approved):

PR #1416, opened 2025-08-04
Fix hanging when searching for KiCad plugins

zampierilucas:
The `ato create` command was hanging on systems with large home directories due to a recursive glob search from `~`. 

**What changed:**
- Check common directories first (`.local`, `.config`, `Documents`, `Library`, `AppData`) 
- Fall back to full home search only if needed
- Same functionality, just faster for most users

**Result:**
- Commands that hung now complete in ~1 second
- Still finds KiCad in unusual locations via fallback


iopapamanoglou:
@sam-mellor please have a look why CI fails

iopapamanoglou (approved):

PR #1417, opened 2025-08-04
Fix: warn on step failure

napowderly:
## Summary
This PR changes the behavior of STEP and GLB 3D model exports to be non-blocking. When these exports fail, the build will now log a warning and continue with the remaining outputs instead of failing entirely.

## Motivation
Currently, if 3D model generation fails (e.g., due to missing 3D models for components), the entire build fails. This is problematic when users need other manufacturing outputs (Gerber files, Pick & Place, BOM, etc.) but don't necessarily need the 3D models.

## Changes
- Modified `generate_manufacturing_data()` to catch `KicadCliExportError` for STEP exports and log a warning instead of failing
- Modified `generate_3d_model()` to catch `KicadCliExportError` for GLB exports and log a warning instead of failing
- Added TODO comments indicating this is a temporary solution until 3D model generation becomes more stable

## Example Output
When a 3D export fails, users will now see:
```
WARNING: Failed to generate STEP file: <error details>. Continuing with remaining manufacturing data generation.
```

Instead of the build failing completely.

## Future Work
As noted in the TODO comments, this error handling should be removed once 3D model generation is stable and reliable. At that point, we'll want builds to fail if 3D exports fail, as it would indicate a real issue that needs to be addressed.

## Breaking Changes
None - this makes the build process more permissive, not less.

### DIFF
--- a/src/atopile/buildutil.py
+++ b/src/atopile/buildutil.py
@@ -416,6 +416,8 @@ def generate_bom(app: Module, solver: Solver) -> None:
 def generate_3d_model(app: Module, solver: Solver) -> None:
     """Generate PCBA 3D model as GLB. Used for 3D preview in extension."""
 
+    # TODO: This try/except should be removed once 3D model generation is stable
+    # For now, we log a warning and continue instead of failing the build
     try:
         export_glb(
             config.build.paths.layout,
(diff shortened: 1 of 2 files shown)


napowderly:
Screenshot 2025-08-04 at 1 40 51 PM

PR #1426, opened 2025-08-06
CI: Disable pytest-history

sam-mellor:
Speeds up test workflow by 500%.

Followup: fix perf issue; consider re-enabling

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -12,7 +12,7 @@ env:
 
 jobs:
   pytest:
-    runs-on: blacksmith-16vcpu-ubuntu-2404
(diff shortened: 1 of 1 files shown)


.github/workflows/pytest.yml
```diff
@@ -12,7 +12,7 @@ env:
 
 jobs:
   pytest:
-    runs-on: blacksmith-8vcpu-ubuntu-2404
+    runs-on: blacksmith-4vcpu-ubuntu-2404
```
sam-mellor, line 15: Not any slower (n=1)

iopapamanoglou (approved):

PR #1429, opened 2025-08-06
Libs: Picker: Move endpoint specification into modules

sam-mellor:
Modules specify API endpoint and expected parameters. Picker no longer needs knowledge of specific modules.

Usage:
```python
@L.rt_field
def pickable(self):
    return F.is_pickable_by_type(
        endpoint=F.is_pickable_by_type.Endpoint.RESISTORS,
        params=[self.resistance, self.max_power, self.max_voltage],
    )
```

### DIFF
--- a/src/faebryk/library/Capacitor.py
+++ b/src/faebryk/library/Capacitor.py
@@ -55,7 +55,7 @@ class Capacitor(Module):
     @L.rt_field
     def pickable(self) -> F.is_pickable_by_type:
         return F.is_pickable_by_type(
-            endpoint="capacitors",
+            endpoint=F.is_pickable_by_type.Endpoint.CAPACITORS,
             params=[self.capacitance, self.max_voltage, self.temperature_coefficient],
         )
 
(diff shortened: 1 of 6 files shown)


src/faebryk/library/Capacitor.py
```diff
@@ -55,12 +55,8 @@ class TemperatureCoefficient(Enum):
     @L.rt_field
     def pickable(self) -> F.is_pickable_by_type:
         return F.is_pickable_by_type(
-            F.is_pickable_by_type.Type.Capacitor,
-            {
-                "capacitance": self.capacitance,
-                "max_voltage": self.max_voltage,
-                "temperature_coefficient": self.temperature_coefficient,
-            },
+            endpoint="capacitors",
```
iopapamanoglou: how is a string better than an enum?

src/faebryk/libs/picker/api/picker_lib.py
```diff
@@ -107,15 +90,17 @@ def _extract_numeric_id(lcsc_id: str) -> int:
 )
 
 
-def _from_smd_size(
-    cls, size: SMDSize, type: F.is_pickable_by_type.Type
-) -> "BackendPackage":
+def _from_smd_size(cls, size: SMDSize, type: type[L.Module]) -> "BackendPackage":
+    if issubclass(type, F.Resistor):
```
sam-mellor, line 94: Only until `package` is a `Parameter`

PR #1447, opened 2025-08-14
Library: Organize into folders, clean variable names, move footprint enum location

nickkrstevski:
Cleaning the standard library:
* Organize into folders for increased discoverability
* Switch to `rated_x` nomenclature for datasheet specifications
* Deprecate old parameter names
* Use new `pickable` syntax
* Move footprint enum into component definition file for increased discoverability


src/faebryk/library/components/TestPoint.py
```diff
@@ -0,0 +1,122 @@
...
+        if self._pad_shape == self.PadShape.SQUARE:
+            pad_size = f"{self._size:.1f}x{self._size:.1f}mm"
+
+        drill_diameter = ""
+        if self._pad_type == self.PadType.THT:
+            drill_diameter = (
+                f"_Drill{supported_pad_sizes_drill_diameters[self._size]}mm"
+            )
+
+        return f"TestPoint_{self._pad_type.value}_{pad_size}{drill_diameter}"
```
iopapamanoglou, line 71: This is a detail of the kicad footprint trait not of the TestPoint module.
Make a new trait class that inherits from the right trait .impl and implement the corresponding abstract function

src/faebryk/library/components/TestPoint.py
```diff
@@ -0,0 +1,122 @@
+# This file is part of the faebryk project
```
iopapamanoglou, line 1: Didn't ruben make a testpoint & mounting hole package?

src/faebryk/library/components/Fuse.py
```diff
@@ -0,0 +1,126 @@
...
+            endpoint=F.is_pickable_by_type.Endpoint.FUSES,
+            params=[
+                self.fuse_type,
+                self.response_type,
+                self.rated_trip_current,
+                self.rated_hold_current,
+                self.rated_voltage,
+                self.rated_power_dissipation,
+            ],
+        )
```
iopapamanoglou, line 78: I don't like this pattern.
Fuse.py/ato should encapsulate everything that is inherent to a Fuse
Being able to be picked by the backend is not an inherent property of Fuse.
Thus the trait should be added at a place that is related to the picking.

src/faebryk/library/components/Capacitor.py
```diff
@@ -0,0 +1,110 @@
...
+        Z5U = auto()
+        X7S = auto()
+        X5R = auto()
+        X6R = auto()
+        X7R = auto()
+        X8R = auto()
+        C0G = auto()
+
+    # TODO: Deprecated unnamed -> terminals
+    terminals = L.list_field(2, F.Electrical)
```
iopapamanoglou, line 35: `electrodes` makes more sense to me (see Diode.anode/cathode)
`terminal` implies some things that I can't put my finger on that I don't think fit to a cap
I would accept having both, and connecting them to each other.

iopapamanoglou (changes requested):
Thanks for the comprehensive cleanup!
Added some comments.

src/faebryk/library/components/Capacitor.py
sam-mellor: I think both is the worst option. We should be opinionated.

iopapamanoglou:

PR #1455, opened 2025-08-16
Exporters: Expose svg (2D) and png (3D) PCB/PCBA renders as build target

ruben-iteng:
## Description

Expose 2D and 3D render exporter as build target. 
Add function to export 3D render via KiCad cli

## Motivation

Who doesn't like auto generated visuals 😌

### DIFF
--- a/src/atopile/targets.py
+++ b/src/atopile/targets.py
@@ -20,11 +20,13 @@ from faebryk.exporters.netlist.netlist import make_fbrk_netlist_from_graph
 from faebryk.exporters.parameters.parameters_to_file import export_parameters_to_file
 from faebryk.exporters.pcb.kicad.artifacts import (
     KicadCliExportError,
+    export_3d_board_render,
     export_dxf,
     export_gerber,
     export_glb,
     export_pick_and_place,
     export_step,
+    export_svg,
     githash_layout,
 )
 from faebryk.exporters.pcb.pick_and_place.jlcpcb import (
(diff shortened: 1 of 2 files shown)


src/atopile/targets.py
```diff
@@ -189,6 +191,35 @@ def generate_3d_models(app: Module, solver: Solver) -> None:
     pass
 
 
+@muster.register(name="3d-image", aliases=["3d-image"], requires_kicad=True)
```
sam-mellor:
```suggestion
@muster.register(name="3d-image", requires_kicad=True)
```

src/atopile/targets.py
```diff
@@ -189,6 +191,35 @@ def generate_3d_models(app: Module, solver: Solver) -> None:
...
+            export_3d_board_render(
+                tmp_layout,
+                image_file=config.build.paths.output_base.with_suffix(".pcba.png"),
+                project_dir=config.build.paths.layout.parent,
+            )
+        except KicadCliExportError as e:
+            raise UserExportError(f"Failed to generate 3D rendered image: {e}") from e
+
+
+@muster.register(name="2d-image", aliases=["2d-image"], requires_kicad=True)
```
sam-mellor:
```suggestion
@muster.register(name="2d-image", requires_kicad=True)
```

ruben-iteng:
> Is there something special about `rpdap_compact`? I'm seeing 3D models in the render for that but not any of the other projects / build targets I've tried.

I also tested another project and that worked fine too. Do your 3D models load in kicad?

Btw 2D image takes <1s to generate 3D >10s

iopapamanoglou (approved):

PR #1456, opened 2025-08-19
Tests: Add a `require_network` marker

Sigmanificient:
## Description

Introduces a new marker for tests that can only be run in a online environment.

## Motivation

This change is meant to simplify the exclusion list on nixpkgs side, which run the tests without network, to ensure reproducibility


iopapamanoglou:
Overlaps with #1436

PR #1460, opened 2025-08-25
CLI: Add package verify command

nickkrstevski:
Adds CLI command to verify package is ready for publishing rather than relying on build/publish workflow.

Package verify command verifies with 2 levels of rigor, standard, and strict.

**Standtard Checks:**
* verify_ato_yaml
* verify_build_exists
* verify_pinned_dependencies
* verify_version_increment

**Strict Checks:**
* All standard checks
* verify_3d_models
* verify_file_structure
* verify_no_warnings
* verify_usage_import
* verify_usage_in_readme

### DIFF
--- a/src/faebryk/libs/picker/lcsc.py
+++ b/src/faebryk/libs/picker/lcsc.py
@@ -428,7 +428,6 @@ class EasyEDAPart:
         return None
 
     def load_datasheet(self):
-        return
         if self.datasheet_url:
             return self.datasheet_url
 


src/atopile/cli/package.py
```diff
@@ -71,6 +73,313 @@ def _apply_version(specd_version: str, config: "Config") -> None:
...
+        if not entry_path.exists():
+            raise UserFileNotFoundError(f"Usage build entry not found: {entry_path}")
+
+        context = bob.index_file(entry_path)
+
+        offending: list[str] = []
+        for _ref, node in context.refs.items():
+            if isinstance(node, Context.ImportPlaceholder):
+                if (
+                    "atopile" not in node.from_path
```
sam-mellor: I think this conflates multiple things that we should separate out:
- a configurable owner allowlist for dependencies
- absolute imports only (leading two parts of import path exists in dependencies)

nickkrstevski: Changed the way we check this to verifying that the import is found in .ato/modules.

If the import starts with "parts/" ignore
If the import does not exist in .ato/modules and does not contain the package identifier then it is offender

src/faebryk/libs/picker/lcsc.py
```diff
@@ -428,6 +428,7 @@ def load_datasheet_for_identifier(
         return None
 
     def load_datasheet(self):
+        return
```
sam-mellor:
```suggestion
```

PR #1462, opened 2025-08-25
Libs: Switch to system cert store

sam-mellor:
Switches from bundled certs (via `certifi`) to native system cert stores for SSL verification, and from `requests` to `httpx`.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = "==3.13.*"
 
 [[package]]
@@ -72,6 +72,7 @@ dependencies = [
     { name = "fastapi-github-oidc" },
     { name = "freetype-py" },
     { name = "gitpython" },
+    { name = "httpx" },
     { name = "kicad-python" },
     { name = "kicadcliwrapper" },
     { name = "matplotlib" },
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -275,6 +276,7 @@ extend-exclude = "atopile/parser/"
 
 [tool.uv.sources]
 atopile-mkdocs-plugin = { path = "tools/atopile_mkdocs_plugin", editable = true }
+atopile-easyeda2kicad = { git = "https://github.com/atopile/easyeda2kicad.py", branch = "feature/system_cert_store" }
```
sam-mellor: Note: release atopile-easyeda2kicad first, then remove this line

src/faebryk/libs/http.py
```diff
@@ -0,0 +1,27 @@
...
+    TimeoutException,
+)
+
+
+@contextlib.contextmanager
+def http_client(
+    headers: dict[str, str] | None = None, verify: bool | ssl.SSLContext = True
+) -> Generator[httpx.Client, None, None]:
+    if verify is True:
+        verify = truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
```
sam-mellor, line 21: Consider adding opt-in to `certifi` via env var so there's a solution for systems without reliable CA bundles available

sam-mellor: Deferred until the `easyeda2kicad` dependency is dropped, since otherwise the behavior won't match

PR #1463, opened 2025-08-26
Libs: Exporters: Generate layout rules

napowderly:
Generate KiCad board rules from atopile/Faebryk design intent so that layout constraints are consistent with the electrical spec. 

Initial scope:

- Differential impedance-driven width and spacing
- Current-driven single-ended trace width
- Differential signal length matching (time skew in ps)
- Voltage driven spacing rules


iopapamanoglou:

PR #1468, opened 2025-08-27
Buildutil: Model as DAG of individual steps

sam-mellor:
Breaks down `buildutil` into a collection of functions with explicitly-modeled dependencies, as part of the same DAG as build output targets. Functionality and UI is unchanged.


src/atopile/build_steps.py
```diff
@@ -0,0 +1,684 @@
...
+    app: Module, solver: Solver, pcb: F.PCB, log_context: LoggingStage
+) -> None:
+    """Generate a Mermaid diagram of the I2C bus tree."""
+    export_i2c_tree(
+        app, solver, config.build.paths.output_base.with_suffix(".i2c_tree.md")
+    )
+
+
+@muster.register(
+    "__default__",
```
iopapamanoglou: Commenting here, because config.py not changed.
It took me a solid minute to find where the stupid defaults are set.
Why is this a magic string?
Make it a const (`DEFAULT_TARGET = "__default__"`) and then import that in config.py and use that.
Then there is a backlink

src/atopile/build_steps.py
```diff
@@ -0,0 +1,684 @@
...
+            self.dependency_dag.add_edge(dep.name, target.name)
+
+        return target
+
+    def register(
+        self,
+        name: str | None = None,
+        aliases: list[str] | None = None,
+        description: str | None = None,
+        virtual: bool = False,
```
iopapamanoglou, line 142: virtual is a stupid name.
seems more like `private` or `dont_expose_to_user` or `not_selectable`

iopapamanoglou:
Is there a cli command to show all targets?
Might be nice

src/atopile/build_steps.py
```diff
@@ -0,0 +1,684 @@
...
+                target.implicit = False
+
+        return [self.targets[name] for name in sorted_names if name in self.targets]
+
+
+muster = Muster()
+
+
+@muster.register("post-design-checks", description="Running post-design checks")
+def post_design_checks(
```
iopapamanoglou: we should have a look into this, to make it non-required

iopapamanoglou (approved):
Nice PR, some good ol' tech debt cleanup!
Added some improvement suggestions.
Those can go into a new PR, that's why I'm approving this boi
You can choose which to already do and which comments to convert into an issue.

src/atopile/build_steps.py
sam-mellor: Bus param resolution should be a separate step.

src/atopile/build_steps.py
sam-mellor: They're selectable and not private, just no-ops. GPT-5 likes `virtual`.

PR #1497, opened 2025-09-10
Libs: Parts: hide value text on fab layer

napowderly:
Remove value text on fab layer because its ugly
 Eg: "1µF ±10% 25V X5R"

### DIFF
--- a/src/faebryk/libs/part_lifecycle.py
+++ b/src/faebryk/libs/part_lifecycle.py
@@ -650,8 +650,8 @@ class PartLifecycle:
                         )
                         pcb_fp.propertys[prop_name].value = prop_value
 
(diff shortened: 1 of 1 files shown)


src/faebryk/libs/part_lifecycle.py
```diff
@@ -650,6 +650,10 @@ def _get_prop_uuid(name: str) -> str | None:
                         )
                         pcb_fp.propertys[prop_name].value = prop_value
 
+                    # Always hide Value properties to prevent fab layer text
+                    if prop_name == "Value":
```
sam-mellor: Check for `prop.layer == "F.Fab"` instead?

PR #1501, opened 2025-09-12
CLI: Add package report-status command

nickkrstevski:
Add report-status CLI command that takes pass fail argument and uses github auth similar to publish cmd.

* Reports some basic information about the computer running the build
* Reports atopile version used to build
* Reports build success bool param
* Reports built_at which is really the reported time.


iopapamanoglou:

PR #1505, opened 2025-09-16
Library: Fix: add back RS485 half duplex

ruben-iteng:
## Description

Add back RS485HalfDuplex that was deleted in https://github.com/atopile/atopile/pull/1488
Also add suggested net name, an example to bring it up to standard, and connected reference

## Motivation

I'm using it in almost all my projects :sweat_smile:


ruben-iteng:
fixed in #1528

PR #1518, opened 2025-10-06
Core: Fix pad primitives

napowderly:
Description of issue:
Custom footprint geometries not supported in zig schema causing some footprints to be malformed.

Changes:
- Extend the Zig KiCad schema (sexp/kicad/pcb.zig) to model custom pad primitives by adding a PadPrimitives struct (with the canonical gr_poly list today) and the PadPrimitive union. Pads now surface an optional primitives field with the correct S-expression metadata, so custom geometry survives load/save on the staging branch.
- Update the generated Python stubs (faebryk/core/zig/pcb.pyi) to expose Pad.primitives, keeping the Python-facing API aligned with the Zig schema

Testing:
Re-installed LCSC part C19190416 via ato add on stage/0.14.x and compared the resulting footprint against the main-branch export—custom pads now retain the gr_poly point lists exactly.

Screenshot 2025-10-06 at 3 48 59 PM
Screenshot 2025-10-06 at 3 48 35 PM

### DIFF
--- a/src/faebryk/core/zig/src/sexp/kicad/pcb.zig
+++ b/src/faebryk/core/zig/src/sexp/kicad/pcb.zig
@@ -524,6 +524,18 @@ pub const FpText = struct {
 // Pad structures
 pub const Drill = f64;
 
+pub const PadPrimitive = union(enum) {
+    gr_poly: Polygon,
+};
+
+pub const PadPrimitives = struct {
+    gr_polys: list(Polygon) = .{},
+
+    pub const fields_meta = .{
+        .gr_polys = structure.SexpField{ .multidict = true, .sexp_name = "gr_poly" },
+    };
+};
+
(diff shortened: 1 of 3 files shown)


src/faebryk/core/zig/src/sexp/kicad/pcb.zig
```diff
@@ -524,6 +524,18 @@ pub const FpText = struct {
 // Pad structures
 pub const Drill = f64;
 
+pub const PadPrimitive = union(enum) {
```
iopapamanoglou: Is this used?

iopapamanoglou (changes requested):
- The union looks sketchy but on quick glance unused.
- not sure whether this requires some additions in pcb transformer and layout reuse where we do some geometry operations on different objects 
- rest looks good

PR #1519, opened 2025-10-06
Libs: Thermal vias dont have net in footprint

napowderly:
Thermal vias in footprints do not have a net assigned. Unfortunately the EasyEDA data does not seem to include a net, so we need to infer it by which pads the via is within.


iopapamanoglou:

Issue #1509, opened 2025-09-23
Component signal/pin assignments can clash


raytallen:
### Current Behaviour

An auto-generated package (in this case a DDR3 BGA), used the same names for address line signals (A0-A13) as well as the BGA pins (A1-A9). 

<img width="396" height="562" alt="Image" src="https://github.com/user-attachments/assets/b69f3386-5ab7-45f7-bce9-df6e1f429b12" />

<img width="460" height="594" alt="Image" src="https://github.com/user-attachments/assets/09c8eb40-b1f4-429b-aac9-e3941314d15e" />

This results in nets getting assigned to the wrong pin locations. For example locations A8 and A9 are getting mapped to the address lines when they should be power and ground respectively

<img width="1302" height="1462" alt="Image" src="https://github.com/user-attachments/assets/6d22966a-52b5-42c8-b784-33f66134b3eb" />

<img width="1760" height="1330" alt="Image" src="https://github.com/user-attachments/assets/3272e921-59ca-47af-beca-4aa3812fcc07" />

### Expected Behaviour

1. Do not allow auto-generated parts to use clashing signal and pin names
2. At build time raise a warning or error that there are duplicate definitions for nodes of the same name

### Possible Solution

#pr number

### Version

0.11.1

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ruben-iteng:
Fixed in #1536

PR #1528, opened 2025-10-23
Core: Rework part2

iopapamanoglou:

PR #1536, opened 2025-11-06
libs: AtoPart add prefix to signal-pin name intesection

ruben-iteng:
## Description

There was a possibility both pin and signal names of a ato component are the same. As the pin name cannot be changed (linked to footprint file), the signal name is prefixed with `SIG_`

## Motivation

fixes https://github.com/atopile/atopile/issues/1509
image


ruben-iteng:
merged into #1528

PR #1549, opened 2026-01-16
Libs: Add Package Version Fallback Support

nickkrstevski:
Add package install fallback functionality
Base commit: af9051575f84a9329212499dc8ef1a3d3ba250c2

## Description

When adding package dependency to project, find latest package version that satisfies minimum required ato version.

## Motivation

Current behavior fails to add package if ato version is not compatible with latest package version.


nickkrstevski:
Cherry-picked commits we want from latest main
	•	f0fc8ebd – libs: fix broken deps error message (#1535)
	•	9773bcc9 – Dep: pygls migrate to v2 (#1533)
	•	d8d0d2e8 – VSCE: Fix `stat` on non-existent watched files (#1512)
	•	c898b71a – Library: Support parallel pull resistors in requires_pulls check (#1500)
	•	83ba0f42 – Fix: hide value text on fab layer (#1497)
	•	base - af905157 – Libs: Packages: Fix missing auth header (#1496)

PR #1558, opened 2026-01-26
Compiler: add handling of field-to-field assignment

rubenbaldewsing:
## Description

- Add missing is_assignalble trait for FielRef (fixes trait not found error)
- Add handling of field-to-field assignment (e.g. mod1.voltage = mod2.voltage)

## Motivation

Used in for example the Relay package where a sub module has a parameter defined (not constrained), upper module links the parameter to another sub module it's parameter, and the user constrains the parameter in their app.

### DIFF
--- a/src/atopile/compiler/ast_visitor.py
+++ b/src/atopile/compiler/ast_visitor.py
@@ -1161,6 +1161,37 @@ class ASTVisitor:
                 target_path, assignable.operand
             )
 
+        # Handle field-to-field assignment (e.g., a.voltage = b.voltage)
+        # This creates an "Is" constraint between the two parameters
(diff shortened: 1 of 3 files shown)


src/atopile/compiler/ast_visitor.py
```diff
@@ -1161,6 +1161,37 @@ def visit_Assignment(self, node: AST.Assignment):
                 target_path, assignable.operand
             )
 
+        # Handle field-to-field assignment (e.g., a.voltage = b.voltage)
```
sam-mellor, line 1164: Has this been possible historically? Makes sense as a special case of expression assignment, but also the aliasing behavior seems potentially confusing.

iopapamanoglou:

PR #1562, opened 2026-01-27
fix some early exception logging and live build status updates

raytallen:


src/atopile/cli/build.py
```diff
@@ -154,10 +144,9 @@ def start(self) -> None:
             self.name,
         ]
 
-        # Copy environment and set IPC channels for worker subprocess
+        # Copy environment and set worker mode
         env = os.environ.copy()
-        if self._event_wfd is not None:
-            env["ATO_BUILD_EVENT_FD"] = str(self._event_wfd)
+        env["ATO_BUILD_WORKER"] = "1"
```
rubenbaldewsing, line 149: Use `ConfigFlagDef` from `src/atopile/config_flags.py` for env vars if possible

raytallen: not a config flag, used as a hack for IPC for now

PR #1569, opened 2026-01-27
Fix: Kicad, 3D and Layout buttons in UI

napowderly:
Buttons now properly trigger opening the UI panels for 3D and Layout as well as open the kicad file. 
Layout/3D panels update to show the active Project/build selected in the UI.
Screenshot 2026-01-28 at 8 42 49 AM


ruben-iteng (approved):
Works for me 👍

PR #1574, opened 2026-01-28
Fix: build/target terminology

napowderly:
We are using Target in alot of places where we mean Build.
Updated example, now builds are called builds:
Screenshot 2026-01-28 at 1 05 01 PM


nickkrstevski (approved):
'target' reserved for desired build output artifacts like step, gerber, glb, etc. makes sense.

PR #1590, opened 2026-01-29
Logging: log configuration errors

nickkrstevski:
When using the GUI, configuration errors were happening before logging was initialized. Early failures had null timestamp so they didnt show up in the gui build queue as failed.


raytallen:
Fix with #1593

PR #1599, opened 2026-01-30
Feature/UI tuning

nickkrstevski:
Front end changes to enforce consistency across style and formatting.

Tab container auto expands to fill space, and shrinks to minimum size then become scrollable:
image
image

Search bars standardized and autoselected upon entering tab
image
image

Structure panel color and layout simplified:
image

Error pages rigidly standardized:
image

Remove build all button, replace with big build button. Remove shading behind dropdowns. REMOVE CAPITAL A FROM atopile:
image

Removed colors from most things, including packages buttons:
image


ruben-iteng:
might want to use `src/ui-server/src/utils/searchUtils.ts` for the searchbar so it has support for regex search 😎

PR #1606, opened 2026-01-30
Library: Add .required for parameters

rubenbaldewsing:
```ato
module App:
  voltage: voltage
  voltage.required = True
  assert voltage within 3.3V +/- 10% # <--- must be set
```

TODO:
- does the logic makes any sense 🤔 
- tests for all types


iopapamanoglou:

PR #1612, opened 2026-01-30
UI: Add sync buttons that appear when packages are out of date

ruben-iteng:
Screenshot 2026-01-31 at 13 18 36

Screenshot 2026-01-31 at 13 19 31


napowderly:
I think we should use little warning popup we have in a few other places that has the message + a button. Dont want a new UI style for it.

PR #1639, opened 2026-02-01
VSCE: add "open ato shell" option to menu

raytallen:
image

image

Works on Windows too!
image


napowderly (approved):

napowderly (approved):

iopapamanoglou (changes requested):
This wont work

Need to be careful with shell escaping and powershell.
Have a look at the original implementation (before rework).

PR #1648, opened 2026-02-09
CLI: Fix dup and unused import statement detection

ruben-iteng:
## Description

fix false positive triggers

## Motivation

fix


nickkrstevski (approved):
looks good

PR #1655, opened 2026-02-10
Packages: Fix package verify command for CI

nickkrstevski:
* Fix ato package verify -s crashing with os._exit(1) when build history DB doesn't exist in CI (Docker containers are ephemeral, so the DB created during ato build is lost before ato package verify runs)
* Add FBRK_LOG_DIR env var to get_log_dir() so CI can persist log DBs to a mounted volume between Docker runs
* verify_no_warnings_or_errors now checks actual log entries for both warnings and errors (previously only checked warning counts from build metadata)
* BuildHistory.get_all() raises instead of os._exit(1), letting callers handle errors gracefully
* Console log filter raised from INFO to WARNING

### DIFF
--- a/src/atopile/logging.py
+++ b/src/atopile/logging.py
@@ -1104,8 +1104,8 @@ class LogHandler(RichHandler):
         #     ):
         #         return False
 
-        # Console only shows WARNING and above
-        if record.levelno < logging.WARNING:
+        # Console only shows INFO and above
+        if record.levelno < logging.INFO:
             return False
 
         return True


src/atopile/logging.py
```diff
@@ -1105,7 +1105,7 @@ def _should_log_to_console(self, record: logging.LogRecord) -> bool:
         #         return False
 
         # Console only shows INFO and above
```
rubenbaldewsing: now WARNING and above

src/atopile/model/sqlite.py
```diff
@@ -223,20 +223,14 @@ def get(build_id: str) -> Build | None:
...
-            )
-            os._exit(1)
+        """Get recent builds. Raises on error so callers can handle gracefully."""
+        with _get_connection(BUILD_HISTORY_DB) as conn:
+            conn.row_factory = sqlite3.Row
+            rows = conn.execute(
+                "SELECT * FROM build_history ORDER BY started_at DESC LIMIT ?",
+                (limit,),
+            ).fetchall()
+            return [BuildHistory._from_row(r) for r in rows]
```
rubenbaldewsing: we should keep the try-except and logging message, just not exit + raise the exception instead ?

ruben-iteng (commented):
Almost ok I think

PR #1657, opened 2026-02-10
VSCE: fix ssh port forwarding

raytallen:


src/vscode-atopile/src/providers/LogViewerProvider.ts
```diff
@@ -70,34 +49,29 @@ export class LogViewerProvider implements vscode.WebviewViewProvider {
...
 
-    if (isDev) {
-      this._view.webview.html = this._getDevHtml();
-    } else {
-      this._view.webview.html = this._getProdHtml(this._view.webview);
-    }
+    this._view.webview.options = createWebviewOptions({
+      isDev,
+      extensionPath,
+      port: backendServer.port,
```
raytallen, line 69: here is the thing that actually fixes this

iopapamanoglou (approved):

PR #1673, opened 2026-02-12
Core: parallelize datasheet download

nickkrstevski:
Download datasheets in parallel to save time.


src/faebryk/exporters/documentation/datasheets.py
```diff
@@ -82,20 +83,38 @@ def export_datasheets(
...
-            logger.error(f"Failed to download datasheet {filename}: {e}")
-            continue
-        logger.debug(f"Downloaded datasheet {filename}")
+        to_download.append((url, file_path))
+
+    if not to_download:
+        return
+
+    # Download in parallel — I/O-bound, independent tasks
+    max_workers = min(8, len(to_download))
```
rubenbaldewsing, line 105:
```suggestion
    max_workers = min(os.cpu_count() or 1, len(to_download))
```

PR #1677, opened 2026-02-12
ERC: Add ElectricPower solve voltage

rubenbaldewsing:
## Description

A new check to validate/solve connected ElectricPowers have compatible voltages.

TODO: benchmark, current implementation seems slow...

## Motivation

Otherwise you can connect voltage-incompatible power rails with eachother


iopapamanoglou:

PR #1689, opened 2026-02-13
Core: Refactor logging implementation, allow audience field

raytallen:
- refactor callsites to use of AtoLogger everywhere
- audience field actually useable
- refactor clear distinction between ConsoleLogHandler and DBLogHandler
- logs not associated with a build or test still get saved to DB (unscoped logger)
- remove silent passes for log failures
- add dblogging setup before cli commands run so that all cli logs make it to db
- improve lifecycle management, active loggers auto-flush their logs
- fix periodic log flushing

### Testing
- CI test run
- Local `ato dev test --ci --keep-open` and verify test log viewer works
- Run builds in extension and verify build logs still work


src/atopile/logging.py
```diff
@@ -1175,190 +722,97 @@ def emit(self, record: logging.LogRecord) -> None:
...
-            identifier,
-            test_name=context_filter,
-            levels=log_levels,
-            audience=audience,
-            after_id=after_id,
-            count=max_count,
-            order=order,
+        if active:
+            return active[0]
+        return AtoLogger._get_unscoped(channel="default", stage="")
```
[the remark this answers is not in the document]
raytallen: This is intended, we want to capture all logs

src/atopile/logging.py
```diff
@@ -1175,190 +717,96 @@ def emit(self, record: logging.LogRecord) -> None:
...
+        ]
+        if len(candidates) != 1:
+            raise RuntimeError(
+                "Expected exactly one active DB logging context; "
+                f"found {len(candidates)}."
+            )
+        return candidates[0]
 
+    def emit(self, record: logging.LogRecord) -> None:
+        db_logger = self._resolve_db_target()
```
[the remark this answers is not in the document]
raytallen: Would be caught in regressions

src/atopile/errors.py
```diff
@@ -767,3 +769,233 @@ def log_discord_banner() -> None:
     from atopile.logging import logger
 
     logger.info(DISCORD_BANNER_TEXT)
+
+
+# =============================================================================
+# Traceback Serialization
```
raytallen, line 775: Moved out of `logging.py` to a more appropriate location

src/atopile/logging.py
```diff
@@ -84,766 +77,437 @@ def _is_serving() -> bool:
...
-    def alert(self, msg: object, *args, **kwargs) -> None:
-        """Log at ALERT level (between WARNING and ERROR)."""
-        if self.isEnabledFor(ALERT):
-            self._log(ALERT, msg, args, **kwargs)
-
-
-logging.setLoggerClass(AtoLogger)
-
-# =============================================================================
-# Structured Traceback Extraction
```
raytallen, line 98: Moved into `errors.py` above

src/atopile/logging.py
```diff
@@ -84,766 +93,574 @@ def _is_serving() -> bool:
...
+            self.db_flush()
 
-    def flush(self) -> None:
-        if self._append_chunk is None:
-            return
-        with self._buffer_lock:
-            if not self._buffer:
+    def db_flush(self) -> None:
+        if self._db_writer is None:
+            raise RuntimeError(
```
nickkrstevski, line 383: **Bug: Race condition.** The periodic flush thread can call `db_flush()` concurrently with `_deactivate_logger()`, which sets `_db_writer = None` after flushing. This `raise RuntimeError` will crash the flush thread on a normal lifecycle event.

Same issue in `_write_db_entry()` at line 365. Both should return early instead of raising:

```python
def db_flush(self) -> None:
    if self._db_writer is None:
        return
```

src/atopile/logging.py
```diff
@@ -1181,190 +875,119 @@ def emit(self, record: logging.LogRecord) -> None:
...
+            return AtoLogger._active_build_logger
+
+        return AtoLogger._get_or_create_unscoped_logger()
+
+    def handleError(self, record: logging.LogRecord) -> None:  # noqa: N802
+        """Never swallow DB logging failures."""
+        _exc_type, exc_value, exc_tb = sys.exc_info()
+        if exc_value is None:
+            raise RuntimeError("DBLogHandler failed without an active exception")
+        raise exc_value.with_traceback(exc_tb)
```
nickkrstevski, line 920: **Bug: Logging should never crash the application.** This re-raises any DB write failure (disk full, SQLite locked, schema mismatch), which violates the `logging` module contract. The original silent `pass` was too quiet, but re-raising is too aggressive — a transient SQLite lock will take down the process.

Write to stderr instead (the standard `logging.Handler.handleError` behavior), or just remove the override entirely:

```python
def handleError(self, record: logging.LogRecord) -> None:
    # Use default behavior: print to stderr without crashing
    super().handleError(record)
```

src/atopile/logging.py
```diff
@@ -1181,190 +875,119 @@ def emit(self, record: logging.LogRecord) -> None:
...
-        identifier,
-        stage=context_filter,
-        levels=log_levels,
-        audience=audience,
-        after_id=after_id,
-        count=max_count,
-        order=order,
-    )
 
+        level = _LEVEL_MAP[record.levelno]
```
nickkrstevski, line 931: **Bug: KeyError on non-standard log levels.** Third-party libraries can use custom log levels not in `_LEVEL_MAP`. This should use `.get()` with a fallback:

```python
level = _LEVEL_MAP.get(record.levelno, Log.Level.DEBUG)
```

src/atopile/logging.py
```diff
@@ -84,766 +93,574 @@ def _is_serving() -> bool:
...
+        cls._set_active_loggers(
+            test=test_logger,
+            unscoped=cls._get_or_create_unscoped_logger(),
+        )
+        return test_logger
+
+    @classmethod
+    def set_active_test_name(cls, test: str) -> None:
+        """Set the current test name on the active test logger context."""
+        cls._active_test_logger.stage_or_test_name = test
```
nickkrstevski, line 452: **Bug: Missing None check.** `cls._active_test_logger` can be `None`, causing `AttributeError`. Compare with `set_active_stage` (line 443) which properly guards:

```python
@classmethod
def set_active_test_name(cls, test: str) -> None:
    if cls._active_test_logger is None:
        raise RuntimeError("No active test logging context")
    cls._active_test_logger.stage_or_test_name = test
```

src/atopile/errors.py
```diff
@@ -767,3 +769,222 @@ def log_discord_banner() -> None:
...
+                    else:
+                        rich_repr_parts.append(f"{key}={_format_value(val)}")
+                elif len(item) == 1:
+                    rich_repr_parts.append(_format_value(item[0]))
+            else:
+                rich_repr_parts.append(_format_value(item))
+        result = f"{type_name}({', '.join(rich_repr_parts)})"
+        return result[:max_len] + "..." if len(result) > max_len else result
+
+    result = repr(value)
```
nickkrstevski: **Bug: Lost safety guards from the move.** The original `_get_pretty_repr` in `logging.py` had a `try/except` wrapping the entire function and used `object.__repr__(value)` as fallback to avoid segfaults from zig-backed objects with freed memory during error handling. This version uses bare `repr(value)` which can segfault on freed zig objects, and has no exception guard so failures during traceback extraction will crash the error handler.

Restore the safety:
```python
def _get_pretty_repr(value: object, max_len: int = 200) -> str:
    try:
        # ... existing rich_repr logic ...
        result = object.__repr__(value)  # not repr(value)
        return result[:max_len] + "..." if len(result) > max_len else result
    except Exception:
        return "<unable to represent>"
```

src/atopile/logging.py
nickkrstevski: We want the logs. If logs are not working we want to crash.

src/atopile/logging.py
nickkrstevski: We want to crash in this case.

PR #1695, opened 2026-02-15
UI: chat window + autorouting

napowderly:
- Added a new ato chat panel in the product UI.
  - Added a new agent layer for chat tooling (tool-call based actions from the panel).
  - Added DeepPCB integration into the autolayout flow with:
      - autolayout job orchestration,
      - provider-aware submission/status/fetch paths,
      - candidate handling tied to the new provider path.
  - Added new autolayout config wiring for run behavior (provider, candidate_count, auto_apply, etc.) as part of the new
    provider flow.
  - Added new strict converter/foundation support for DeepPCB-native layout export behavior.
  - Added new datasheet handling behavior updates in the tool/policy path for the agent/autolayout flow.

  Operationally:

  - Both agent panel and DeepPCB flow are opt-in, off by default.
  - They require explicit env toggles + local API keys.
  - Current intended deployment is webIDE + local dev, with broader local use intended to depend on a backend auth server
    path later.

Screenshot 2026-02-22 at 10 39 44 AM


src/atopile/server/agent/skills.py
napowderly, line 1: dont think we want this any more

src/atopile/server/agent/tool_definitions.py
napowderly, line 1: should this be like pydantic instead?

iopapamanoglou:

PR #1710, opened 2026-02-18
KiCad: Fix layout reuse 

raytallen:
Fix KiCad layout reuse mapping for grouped sublayouts

### DIFF
--- a/src/atopile/layout.py
+++ b/src/atopile/layout.py
@@ -209,6 +209,7 @@ def attach_subaddresses_to_modules(app: fabll.Node):
     )
     if not pcb_modules:
         return
+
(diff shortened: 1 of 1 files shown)


src/atopile/layout.py
```diff
@@ -209,17 +209,23 @@ def attach_subaddresses_to_modules(app: fabll.Node):
...
-        )
-        for footprint_child in footprint_children:
-            trait_instance = in_sub_pcb_bound.create_instance(g=g).setup(
-                sub_root_module=module
-            )
-            # Use add_instance_to for trait instances, not add_to
-            fabll.Traits.add_instance_to(footprint_child, trait_instance)
+        for footprint_child in all_footprints:
+            if not footprint_child.is_descendant_of(module):
+                continue
```
iopapamanoglou: that seems pretty slow

PR #1726, opened 2026-02-22
PCB: fix new pcb generation and re-enable ci tests

raytallen:


sam-mellor:
Note test runners now have `kicad-cli` available

PR #1754, opened 2026-03-03
Compiler: Support `pointer_field = new Type` in .ato

rubenbaldewsing:
Enable assigning `new Type` to Python-defined Pointer fields in .ato code, both inside derived modules (`from` inheritance) and via nested paths.


iopapamanoglou:

PR #1759, opened 2026-03-04
CLI: Reduce spam by downgrading info to debug

nickkrstevski:
## Description

Greatly reducing number of random info prints and shifting towards debug so the default cli experience is pleasant.

## Motivation

The CLI looks like shit with all the info prints:
image

CLI looks beautiful without them:
image


iopapamanoglou:

PR #1764, opened 2026-03-04
LSP: Fix crash of LSP + location of squigly lines

rubenbaldewsing:
## Description

No more exceptions or warnings on line 1 column 1, properly display squiggly lines at the source.
                                                           
  - Fix crash by making DeprecatedException extend SourceLocatedUserException and  
  adding getattr guards for other non-source-located exception types               
  - Thread file_path from the LSP through build_source so AST nodes get stamped    
  with their filepath
  - Add a source location context var so deprecation warnings raised anywhere
  during AST visiting (overrides, library code) automatically get the current
  node's location
  - Use from_ctx() in the ANTLR visitor for parser-phase deprecation warnings

## Motivation

  The LSP was crashing with AttributeError: 'DeprecatedException' object has no    
  attribute 'origin_start' because DeprecatedException extended UserException      
  instead of SourceLocatedUserException. Additionally, deprecation warnings always
  displayed at line 1 column 1 because build_source wasn't passing file_path to the
   ANTLR visitor.

### DIFF
--- a/src/atopile/errors.py
+++ b/src/atopile/errors.py
@@ -11,6 +11,7 @@ This module provides:
 from __future__ import annotations
 
 import contextlib
+import contextvars
 import logging
 from abc import ABC, abstractmethod
 from collections.abc import Iterator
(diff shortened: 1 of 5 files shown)


src/atopile/lsp/lsp_server.py
```diff
@@ -217,21 +217,29 @@ def exception_to_diagnostic(
...
             stop_line, stop_col = start_line + 1, 0
 
-    # Convert from 1-indexed (ANTLR) to 0-indexed (LSP)
-    start_line = max(start_line - 1, 0)
-    stop_line = max(stop_line - 1, 0)
+        # Convert from 1-indexed (ANTLR) to 0-indexed (LSP)
+        start_line = max(start_line - 1, 0)
+        stop_line = max(stop_line - 1, 0)
+    elif hasattr(exc, "src_filepath"):
+        # Fallback: direct file location (e.g. from AST nodes without ANTLR tokens)
```
sam-mellor, line 237: How would this happen?

src/atopile/lsp/lsp_server.py
```diff
@@ -245,7 +253,7 @@ def exception_to_diagnostic(
         ),
         message=exc.message,
         severity=severity,
-        code=exc.code,
+        code=getattr(exc, "code", None),
```
sam-mellor, line 256: I don't like all these `getattr`s. Are they just because ANTLR isn't typed properly?

src/atopile/errors.py
```diff
@@ -477,15 +504,31 @@ class UserExportError(SourceLocatedUserException):
     """Raised when there's an error exporting a file via the KiCad CLI."""
 
 
+class DeprecatedException(SourceLocatedUserException):
+    """This feature is deprecated and will be removed in a future version."""
+
+    def __init__(self, msg: str, *args, **kwargs):
+        super().__init__(msg, *args, **kwargs)
+        # Auto-capture source location from context if no ANTLR token is set
```
sam-mellor, line 512: This logic isn't `DeprecatedException`-specific

iopapamanoglou:

PR #1776, opened 2026-03-05
CI: Make package regressions run nightly

raytallen:
## Description

- align the regression workflow setup with the existing pytest workflow
- switch regression to run via `atopile dev test` with a regression-only marker selection
- change regression triggers to manual dispatch, nightly on main, and published releases
- add a 600 second `FBRK_TEST_TEST_TIMEOUT` for regression runs

## Motivation

The regression workflow had drifted from the main pytest workflow and was failing before test collection. This keeps the execution path consistent while moving regression off PR and main push validation into the intended manual, nightly, and release contexts.


nickkrstevski (approved):
Great, works. I suggest more convenient viewing/dashboard for nightly job.

iopapamanoglou:

PR #1778, opened 2026-03-05
Build: Fix TestPyPI deploy versioning

raytallen:
## Description

- strip Hatch VCS local version metadata so TestPyPI receives a publishable PEP 440 version
- make `workflow_dispatch` run the same publish path as `main` pushes in `deploy.yml`

## Motivation

- TestPyPI rejects package versions with local `+...` suffixes, which caused deploy failures on `main`
- manual deploy runs should exercise the same publish path as the automatic `main` workflow


iopapamanoglou:

PR #1784, opened 2026-03-06
Server: file watcher fix

napowderly:
The file watcher is now where it should be: in the Python backend, not in the extension.

  Current state:

  - The extension no longer uses createFileSystemWatcher for sidebar, package explorer, PCB, or 3D model updates.
  - The only real filesystem watcher engine is the shared Python watcher in file_watcher.py.
  - Everything else is a logical subscription on top of that shared watcher.

  How it works now:

  - file_watcher.py
      - owns the shared watchdog observer
      - handles debouncing, filtering, hash suppression, and watch root reuse
  - project_files.py
      - subscribes through the shared watcher
      - rescans the selected project tree
      - emits project_files_changed
  - resource_files.py
      - subscribes through the same shared watcher
      - watches current PCB / 3D output files
      - emits resource_file_changed
  - actions.py
      - exposes watchProjectFiles
      - exposes watchResourceFile
      - exposes getPackageDependencyStatus

  On the extension side:

  - backendServer.ts now keeps one extension-host WebSocket connection to the backend for events and action responses.
  - pcb.ts, 3dmodel.ts, and packagexplorer.ts consume backend events instead of watching files locally.
  - file-resource-watcher.ts is gone.

  So the practical summary is:

  - one backend watcher engine
  - zero extension-owned filesystem watchers
  - sidebar, package status, PCB preview, and 3D preview all update from backend-owned file watching

  That is a materially cleaner architecture than what main had before, because there is now a single ownership boundary for
  filesystem change detection.


iopapamanoglou:

PR #1800, opened 2026-03-11
Solver: Fix Arithmetic Infinite Loop Edge Case

nickkrstevski:
This pattern caused failure in ti-lv2842x: package-update-0.14.x-ti-lv2842x
Very similar pattern causing failures in: ti-tpsm863257

Symptom: The solver enters an infinite loop when an assert contains arithmetic (e.g. subtraction) involving a variable computed from a product of two parameters (dropout_voltage = max_current * dropout_resistance), but works fine when a literal value is used directly in its place.

### DIFF
--- a/test/compiler/test_runtime.py
+++ b/test/compiler/test_runtime.py
@@ -3010,11 +3010,6 @@ class TestAssertWithComputedVariableInArithmetic:
             "App",
         )
 
-    @pytest.mark.xfail(
-        reason="Solver infinite loop when assert uses computed variable in subtraction",
-        raises=TimeoutError,
-        strict=True,
-    )
     def test_assert_subtraction_computed_variable_with_solver(self):
         """Regression: solver should handle computed variable in assert subtraction.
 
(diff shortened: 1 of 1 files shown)


test/compiler/test_runtime.py
```diff
@@ -2934,6 +2934,191 @@ def test_cross_module_parameter_addition_incompatible(self):
...
+                output_voltage = 0.9V to 6V
+                max_current = 3A
+                dropout_resistance = 69mohm
+                dropout_voltage = max_current * dropout_resistance
+                assert input_voltage >= output_voltage + dropout_voltage
+            """,
+            "App",
+        )
+
+    @pytest.mark.xfail(
```
rubenbaldewsing: don't use xfail. the test itself should pass. instead catch the exception with `with pytest.raises(<exception>)`

rubenbaldewsing (changes requested):

test/compiler/test_runtime.py
nickkrstevski: Good call, removed xfail because after fix is implemented they will pass.

iopapamanoglou:

PR #1802, opened 2026-03-11
TypeGraph: Fix corruption: dangling GraphView pointers & string lifetimes

rubenbaldewsing:
## Description

Clanker found these 2 problems after a user described the issue. @sam-mellor I can't really tell if the problems and fixes makes sense especially for the solver, but it might be worth a quick check (there are also 2 tests to reproduce the problems).

### 1. **String lifetimes:**
Type name strings were `allocator.dupe`'d from the GraphView's arena allocator and stored in the global `DynamicAttributes` array. After destroy, `get_type_name()` would read freed string memory and return garbage. Fixed by using `std.heap.c_allocator` for all 9 string duplication sites in `faebryk_py.zig` — these strings are effectively immortal, matching the lifetime of nodes/edges/attrs.

### 2. **Stale GraphView pointers:**
`BoundNodeReference` holds a `*GraphView` pointer. After destroy, edge lookups (`get_single_edge`, `visit_edges_of_type`) would read freed memory. Fixed by:
  - Adding a `destroyed` flag to `GraphView`, set in `deinit()` before freeing the arena
  - Adding early-return guards in `get_single_edge()` and `visit_edges_of_type()`
  - Stopping `destroy()` from freeing the `GraphView` struct itself (only the arena is freed), so the `destroyed` flag remains
   readable

**Solver rebinding:** Added `Node.rebind(g)` method and used it in `Mutator.get_copy()` / `get_operand_copy()` to ensure nodes already in `G_out` are bound to the correct GraphView.

## Motivation

TypeGraph corruption was causing `isinstance()` failures and garbage type names during solver mutations. The root causes were use-after-free bugs: one from dangling `*GraphView` pointers in `BoundNodeReference`, and another from type name strings allocated on a GraphView's arena but stored globally.

### DIFF
--- a/src/faebryk/core/node.py
+++ b/src/faebryk/core/node.py
@@ -1505,6 +1505,10 @@ class Node[T: NodeAttributes = NodeAttributes](metaclass=NodeMeta):
     def graphs_match(*gs: graph.GraphView) -> bool:
         return len(set(g.get_self_node().node().get_uuid() for g in gs)) == 1
 
+    def rebind(self, g: graph.GraphView) -> Self:
+        """Return the same node bound to a different GraphView."""
+        return self.bind_instance(instance=g.bind(node=self.instance.node()))
+
     def copy_into(self, g: graph.GraphView) -> Self:
         """
         Copy all nodes in hierarchy and edges between them and their types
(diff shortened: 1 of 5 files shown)


iopapamanoglou:

PR #1803, opened 2026-03-11
Library: board shape

napowderly:
Screenshot 2026-03-11 at 12 37 27 PM

Also little bonus bug fix, we were applying has_associated_footprint a second time in setup_from_pads_and_leads so a little ~2x speedup on pcb load


examples/quickstart/quickstart.ato
```diff
@@ -1,5 +1,11 @@
+import RectangularBoardShape
 import Resistor
 
 module App:
+    board = new RectangularBoardShape
```
nickkrstevski, line 5: The kicad layout for quickstart does not contain this board shape, needs a build and commit

nickkrstevski (approved):
Looks good, works.

nickkrstevski:
Would be cool to add parametric mounting holes as an option so you could truly be done with board shape for simple boards.

iopapamanoglou:

PR #1835, opened 2026-08-07
add resistance alias for ohm

### LINKED ISSUE #1806: `Unit not found: resistance` – missing alias in Unit registry
### Current Behaviour

When using `resistance` as a type annotation in `.ato` files (e.g. `total_resistance: resistance`), the build fails with the output below. I noted this bug when trying out the example from https://docs.atopile.io/atopile-0.14.x/essentials/2-the-ato-compiler.

```
ValueError
Unit not found: resistance
```

I haven't dug into the code myself, but Claude Code pointed me to the issue and the following analysis:

The unit symbol registry in `faebryk/library/Units.py` (line ~1214) defines:

```python
_UnitRegistry.Ohm: ("Ω", "ohm", "ohms"),
```

Other units include their "physical quantity" name as an alias – for example:
- `Volt`: `("V", "volt", "voltage")`
- `Ampere`: `("A", "ampere", "current")`

But `Ohm` is missing `"resistance"` as an alias. This causes `decode_symbol()` to raise `UnitNotFoundError` when encountering `resistance` in a declaration.

To reproduce:

```ato
module Example:
    total_resistance: resistance
```

```
$ ato build
ValueError
Unit not found: resistance
```

Environment:
- atopile v0.14.1005 (installed via `uv tool`)
- faebryk bundled version (same package)
- Distributor ID: Ubuntu
- Description:    Ubuntu 24.04.4 LTS
- Release:        24.04
- Codename:       noble

### Expected Behaviour

As stated above resistance seems to be a valid indentifier vor a parameter. So the error indicates otherwise. A workaround is to specify a parameter with:
```ato
module Example:
    total_resistance: ohm
```
insteade
```ato
module Example:
    total_resistance: resistance
```

### Possible Solution

Expected fix in `faebryk/library/Units.py` (line ~1214):

```python
_UnitRegistry.Ohm: ("Ω", "ohm", "ohms", "resistance"),
```

### Version

0.14.1005

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

harshverma27:
Fixes #1806

Added `"resistance"` to the Ohm entry in `_UNIT_SYMBOLS`, so `resistance` decodes as a unit the way `voltage` and `current` already do for Volt and Ampere.

### DIFF
--- a/src/faebryk/library/Units.py
+++ b/src/faebryk/library/Units.py
@@ -1211,7 +1211,7 @@ _UNIT_SYMBOLS: dict[_UnitRegistry, tuple[str, ...]] = {
     _UnitRegistry.Coulomb: ("C",),
     _UnitRegistry.Volt: ("V", "volt", "voltage"),
     _UnitRegistry.Farad: ("F", "farad"),
-    _UnitRegistry.Ohm: ("Ω", "ohm", "ohms"),
+    _UnitRegistry.Ohm: ("Ω", "ohm", "ohms", "resistance"),
     _UnitRegistry.Siemens: ("S",),
     _UnitRegistry.Weber: ("Wb",),
     _UnitRegistry.Tesla: ("T",),


harshverma27:
@iopapamanoglou mind reviewing the change?

